PR #6, opened 2020-01-16
Removed deprecated options

g4s8:
#1 - removed deprecated JDK option `MaxPermSize`

> Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0

`mx1g` also not needed here, since Java heap size will set to the maximum available memory on the running machine. This options usually used to reduce the size of available memory.


yegor256:
@rultor merge

PR #8, opened 2020-01-20
Fixed rultor release commands

g4s8:
Rultor has broken `$HOME/.m2/settings.xml`, using `../settings.xml` explicitly.


yegor256:
@rultor release, tag is `0.3.0`

PR #9, opened 2020-01-20
Use Java 9 Flow for data I/O

Sammers21:


g4s8:
@Sammers21 please fix CI build first

Sammers21:
@g4s8, done

src/main/java/com/artipie/asto/Simple.java
```diff
@@ -0,0 +1,157 @@
...
+                            prefix
+                        )
+                    );
+                }
+                final Path path = Paths.get(this.dir.toString(), prefix);
+                final Collection<String> keys = Files.walk(path)
+                    .filter(Files::isRegularFile)
+                    .map(Path::toString)
+                    .map(
+                        p -> p.substring(
```
g4s8: @Sammers21 what does this code do? I'd move it to new method with readable name and javadoc

src/main/java/com/artipie/asto/Simple.java
```diff
@@ -0,0 +1,157 @@
...
+                );
+                return keys;
+            }).to(SingleInterop.get());
+    }
+
+    @Override
+    public CompletableFuture<Void> save(final String key, final Flow.Publisher<Byte> content) {
+        final Completable result = Flowable.fromPublisher(FlowAdapters.toPublisher(content))
+            .toList()
+            .map(bytes -> bytes.toArray(new Byte[0]))
```
g4s8: @Sammers21 these two maps can be converted into single one

src/main/java/com/artipie/asto/Simple.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.Flow;
+import java.util.stream.Collectors;
+import org.reactivestreams.FlowAdapters;
+
+/**
+ * Simple storage, in files.
+ *
+ * @since 0.1
+ */
+public final class Simple implements Storage {
```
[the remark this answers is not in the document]
Sammers21: Make sense.  Renamed to `FileStorage`

src/main/java/com/artipie/asto/Simple.java
```diff
@@ -0,0 +1,157 @@
...
+     */
+    public Simple(final Path path) {
+        this.dir = path;
+    }
+
+    @Override
+    public CompletableFuture<Boolean> exists(final String key) {
+        return (CompletableFuture<Boolean>) Single.fromCallable(
+            () -> {
+                final Path path = Paths.get(this.dir.toString(), key);
```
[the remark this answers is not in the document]
Sammers21: No, it can't. It is used 3 times later.

src/main/java/com/artipie/asto/Simple.java
```diff
@@ -0,0 +1,157 @@
...
+                return Files.exists(path);
+            }).to(SingleInterop.get());
+    }
+
+    @Override
+    public CompletableFuture<Collection<String>> list(final String prefix) {
+        return (CompletableFuture<Collection<String>>) Single.fromCallable(
+            () -> {
+                if (!prefix.endsWith("/")) {
```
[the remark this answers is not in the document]
Sammers21: That is how directories are specified

src/main/java/com/artipie/asto/Simple.java
Sammers21: Yes, but should they?

src/main/java/com/artipie/asto/Simple.java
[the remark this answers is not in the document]
g4s8: @Sammers21 the whole code block here is:
```java
() -> {
  final Path path = Paths.get(this.dir.toString(), key);
  return Files.exists(path);
}
```
I mean it can be:
```java
() -> Files.exists(Paths.get(this.dir.toString(), key))
```

src/main/java/com/artipie/asto/Simple.java
g4s8: @Sammers21 I think yes in this case, since both maps exists here for one purpose only: convert list of bytes `ByteArray` class.

Sammers21: Agree

src/main/java/com/artipie/asto/FileStorage.java
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * @param path The path to the dir
+     */
+    public FileStorage(final Path path) {
+        this.dir = path;
+    }
+
+    @Override
+    public CompletableFuture<Boolean> exists(final String key) {
+        return (CompletableFuture<Boolean>) Single.fromCallable(
```
g4s8: @Sammers21 I don't think we need to cast `CompletionStage` to `CompletableFuture`, since stage interface has `CompletableFuture<T> toCompletableFuture()` method which returns self in case if the stage is a future or transforms itself into future if it is not. See https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html#toCompletableFuture--

So it would be OK to return `CompletionStage` here.

src/main/java/com/artipie/asto/FileStorage.java
```diff
@@ -0,0 +1,143 @@
...
+            () -> Files.exists(Paths.get(this.dir.toString(), key))
+        ).to(SingleInterop.get());
+    }
+
+    @Override
+    public CompletableFuture<Collection<String>> list(final String prefix) {
+        return (CompletableFuture<Collection<String>>) Single.fromCallable(
```
g4s8: @Sammers21 same here: we don't need this cast

src/main/java/com/artipie/asto/FileStorage.java
```diff
@@ -0,0 +1,143 @@
...
+                            final Path target = Paths.get(this.dir.toString(), key);
+                            target.getParent().toFile().mkdirs();
+                            Files.write(target, bytes, StandardOpenOption.CREATE_NEW);
+                            Logger.info(
+                                this,
+                                "Saved %d bytes to %s: %s",
+                                bytes, key, target
+                            );
+                        }));
+        return ((CompletableFuture<Object>) result.to(CompletableInterop.await()))
```
g4s8: @Sammers21 same here: we don't need this cast

src/main/java/com/artipie/asto/FileStorage.java
```diff
@@ -0,0 +1,143 @@
...
+                    "Found %d objects by the prefix \"%s\" in %s by %s: %s",
+                    keys.size(), prefix, this.dir, path, keys
+                );
+                return keys;
+            }).to(SingleInterop.get());
+    }
+
+    @Override
+    public CompletableFuture<Void> save(final String key, final Flow.Publisher<Byte> content) {
+        final Completable result = Flowable.fromPublisher(FlowAdapters.toPublisher(content))
```
g4s8: @Sammers21 the result can be inlined like in all other methods: `return Flowable.fromPublisher(...)`

src/main/java/com/artipie/asto/FileStorage.java
```diff
@@ -0,0 +1,143 @@
...
+                                "Saved %d bytes to %s: %s",
+                                bytes, key, target
+                            );
+                        }));
+        return ((CompletableFuture<Object>) result.to(CompletableInterop.await()))
+            .thenApply(o -> null);
+    }
+
+    @Override
+    public CompletableFuture<Flow.Publisher<Byte>> value(final String key) {
```
g4s8: @Sammers21 same here: we can return `CompletionStage`

Sammers21:
@g4s8, I'm against the idea of returning `CompletionStage` instead of `CompletableFuture`.  It makes API more complex to use.

yegor256:
@rultor merge

Sammers21:
@rultor merge

g4s8:
@rultor merge

yegor256:
@rultor merge

Sammers21:
@rultor merge

Sammers21:
@rultor release, tag is 0.4.0

Sammers21:
@rultor release, tag is `0.4.0`

g4s8:
@rultor release, tag=`0.4.0`

g4s8:
@rultor release, tag=`0.4.0`

g4s8:
@rultor release, tag=`0.4.0`

g4s8:
@rultor release, tag=`0.4.0`

g4s8:
@rultor release, tag=`0.4.0`

g4s8:
@rultor release, tag=`0.4.0`

g4s8:
@rultor release, tag=`0.4.0`

g4s8:
@rultor release, tag=`0.4.0`

g4s8:
@rultor release, tag=`0.4.0`

Issue #19, opened 2020-01-24
Abstract key interface


g4s8:
Use abstract key interface instead of `String` which is used as Unix path now.
Related: https://github.com/artipie/docker-adapter/pull/33 https://github.com/artipie/docker-adapter/pull/34

g4s8:
@rultor release, tag=`0.6.0`

PR #20, opened 2020-01-24
Abstract key

g4s8:
#19 - using abstract `Key` interface instead of `String`

### DIFF
--- a/src/main/java/com/artipie/asto/Key.java
+++ b/src/main/java/com/artipie/asto/Key.java
@@ -1,0 +1,91 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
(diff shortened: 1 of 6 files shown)


src/main/java/com/artipie/asto/Key.java
```diff
@@ -0,0 +1,91 @@
...
+         * Parts.
+         */
+        private final Iterable<String> parts;
+
+        /**
+         * Ctor.
+         * @param parts Parts
+         */
+        public From(final String... parts) {
+            this(Arrays.asList(parts));
```
Sammers21: Can you make the constructors code-free?

g4s8: @Sammers21 should be fixed in https://github.com/artipie/asto/pull/20/commits/5adcf2b2ac85bc3b765d35931ad50ec760fe8a03

PR #22, opened 2020-01-24
expose ByteArray to public

### LINKED ISSUE #21: ByteArray should be exposed to public
It is useful for adapters

Sammers21:
fixes #21

### DIFF
--- a/src/main/java/com/artipie/asto/ByteArray.java
+++ b/src/main/java/com/artipie/asto/ByteArray.java
@@ -23,13 +23,15 @@
  */
 package com.artipie.asto;
 
+import java.util.List;
+
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/com/artipie/asto/ByteArray.java
```diff
@@ -23,25 +23,36 @@
...
+
 /**
  * Byte array wrapper with ability to transform it to
  * boxed and primitive array.
  *
- * @since 0.1
+ * @since 0.7
  */
-class ByteArray {
+public class ByteArray {
```
g4s8: @Sammers21 I think it should be `final`

Sammers21:
@rultor , release tag=`0.8.0`

PR #34, opened 2020-01-31
Fixed safe copying in Key.From constructor

olegmoz:
Added protective copying of key parts to Key.From constructor.


Sammers21:
@rultor, release tag=`0.11.1`

amihaiemil:
@Sammers21 next time, it is better to open an Issue for the release, and state in that Issue what does the new release contain....

PR #103, opened 2020-03-10
Remove RxFile(Path) constructor

Sammers21:
In general, it is a bad idea to use this constructor in production code, because it creates 2xCPU amount of threads.

### DIFF
--- a/src/main/java/com/artipie/asto/fs/RxFile.java
+++ b/src/main/java/com/artipie/asto/fs/RxFile.java
@@ -60,14 +60,6 @@ public class RxFile {
     /**
      * Ctor.
      * @param file The wrapped file.
-     */
-    public RxFile(final Path file) {
-        this(file, Vertx.vertx().fileSystem());
-    }
-
-    /**
-     * Ctor.
-     * @param file The wrapped file.
      * @param fls The file system.
      */
     public RxFile(final Path file, final FileSystem fls) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/artipie/asto/fs/RxFile.java
```diff
@@ -57,14 +56,6 @@
      */
     private final FileSystem fls;
 
-    /**
-     * Ctor.
-     * @param file The wrapped file.
-     */
-    public RxFile(final Path file) {
```
g4s8: @Sammers21 maybe we can use some default Vertx FileSystem for default constructor to avoid writing same code in all adapters? E.g.:
```java
class RxFile {
  private static final FileSystem FS_DEFAULT = Vertx.fileSystem();

  public RxFile(final Path file) {
    this(file, FS_DEFAULT);
  }
}
```

Sammers21: @g4s8 , the code you suggest won't compile.  The other thing is that we need to ensure that a vert.x instance is created per JVM. i.e. we need to initialize it only once.

Sammers21:
`FileSystem(Path)` also have been removed

g4s8:
@rultor release, tag=`0.15`

0crat:
Job `gh:artipie/asto#103` is not assigned, can't get performer

0crat:
Job `gh:artipie/asto#103` doesn't exist in WBS, can't create order

Issue #131, opened 2020-03-25
FileStorage incorrectly processes list(Key.ROOT) call


brastak:
Calculation of path prefixes works incorrectly when Key.ROOT is passed to the list method.

Let's assume:
- base folder is /storage
- artifact is persisted in /storage/artifact.bin

When FileStorage.list() is called for the ROOT, than prefix length is 8 ("/storage".length() + "".length()) and the rest part of the path is /artifact.bin. We can not build key for this string and method fails (it splits to ["", "artifact.bin"] array and key can not contain empty parts).

So we need either support Key construction from the string with the leading "/" or change FileStorage to handle Key.ROOT a special way

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8) please, pay attention to this issue

g4s8:
@rultor release, tag=`0.16.1`

0crat:
Job `gh:artipie/asto#131` is not assigned, can't get performer

PR #133, opened 2020-03-25
fix FileStorage:list with Key.ROOT issue: add missed file

brastak:
missed file for #131 
previous version still works (0.16.1) but just because of using Key.ROOT singleton in the code. Call with Key.From(Collections.emptyList()) will fails without this fix


0crat:
Job `gh:artipie/asto#133` is not assigned, can't get performer

PR #170, opened 2020-05-13
Add toString for Key

Sammers21:
This change would address a number of logging issues when #string method is not called.

### DIFF
--- a/src/main/java/com/artipie/asto/Key.java
+++ b/src/main/java/com/artipie/asto/Key.java
@@ -85,16 +85,6 @@ public interface Key {
         }
 
         @Override
-        public boolean equals(final Object another) {
-            return this.origin.equals(another);
-        }
-
-        @Override
-        public int hashCode() {
-            return this.origin.hashCode();
-        }
-
-        @Override
         public final String toString() {
             return this.string();
         }
--- a/src/main/java/com/artipie/asto/Key.java
+++ b/src/main/java/com/artipie/asto/Key.java
@@ -83,6 +83,21 @@ public interface Key {
         public Optional<Key> parent() {
             return this.origin.parent();
         }
+
+        @Override
+        public boolean equals(final Object another) {
+            return this.origin.equals(another);
+        }
+
+        @Override
+        public int hashCode() {
+            return this.origin.hashCode();
+        }
+
+        @Override
+        public final String toString() {
+            return this.string();
+        }
     }
 
     /**
@@ -187,6 +202,11 @@ public interface Key {
         @Override
         public int hashCode() {
             return Objects.hash(this.parts);
+        }
+
+        @Override
+        public String toString() {
+            return this.string();
         }
     }
 }
--- a/src/main/java/com/artipie/asto/fs/FileStorage.java
+++ b/src/main/java/com/artipie/asto/fs/FileStorage.java
@@ -113,7 +113,7 @@ public final class FileStorage implements Storage {
                 Logger.info(
                     this,
                     "Found %d objects by the prefix \"%s\" in %s by %s: %s",
-                    keys.size(), prefix, this.dir, path, keys
+                    keys.size(), prefix.string(), this.dir, path, keys
                 );
                 return keys;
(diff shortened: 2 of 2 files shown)


Sammers21:
@g4s8, can you pay attention to the Pull Request?

src/main/java/com/artipie/asto/Key.java
```diff
@@ -83,6 +83,21 @@ public final String string() {
         public Optional<Key> parent() {
             return this.origin.parent();
         }
+
+        @Override
+        public boolean equals(final Object another) {
+            return this.origin.equals(another);
```
g4s8, line 89: @Sammers21 this implementation is not correct, according to `equals` specification of `java.lang.Object` :
 1. It should be "reflexive", but now `key.equals(key)` will fall into stack overflow error
 2. It should be "symmetric", but it is not: `wrap.equals(origin) != origin.equals(wrap)`

[See the docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)):
>  The equals method implements an equivalence relation on non-null object references:
>
>    It is reflexive: for any non-null reference value x, x.equals(x) should return true.
>    It is symmetric: for any non-null reference values x and y, x.equals(y) should return true if and only if y.equals(x) returns true.
>    It is transitive: for any non-null reference values x, y, and z, if x.equals(y) returns true and y.equals(z) returns true, then x.equals(z) should return true.
>    It is consistent: for any non-null reference values x and y, multiple invocations of x.equals(y) consistently return true or consistently return false, provided no information used in equals comparisons on the objects is modified.
>    For any non-null reference value x, x.equals(null) should return false. 
>
> The equals method for class Object implements the most discriminating possible equivalence relation on objects;  that is, for any non-null reference values x and y, this method returns true if and only if x and y refer to the same object (x == y has the value true).

Note that it is generally necessary to override the hashCode method whenever this method is overridden, so as to maintain the general contract for the hashCode method, which states that equal objects must have equal hash codes.

src/main/java/com/artipie/asto/Key.java
```diff
@@ -83,6 +83,21 @@ public final String string() {
...
+        }
+
+        @Override
+        public int hashCode() {
+            return this.origin.hashCode();
```
g4s8, line 94: @Sammers21 there should be additional shift for origins hash code to avoid mixing of origin and wrap in hash maps:`return 17 + origin.hashCode()`

0crat:
This pull request [#170](https://github.com/artipie/asto/issues/170) is assigned to @paulodamaso[/z](https://www.0crat.com/u/paulodamaso), here is [why](https://www.0crat.com/footprint/CT2E6TK9B/41ee5716-e8de-452c-a57d-c6bd0ccd319d); the budget is 15 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§27](http://www.zerocracy.com/policy.html#27) and when you decide to accept the changes, inform @g4s8[/z](https://www.0crat.com/u/g4s8) (the architect) right in this ticket; if you decide that this PR should not be accepted ever, also inform the architect; this [blog post](http://www.yegor256.com/2015/02/09/serious-code-reviewer.html) will help you understand what is expected from a code reviewer; there will be a monetary reward for this job

paulodamaso:
@Sammers21 I don't have any additional comments, just fix the problems pointed by @g4s8 and add tests to them

0crat:
Job was finished in 22 hours, bonus for fast delivery is possible (see [§36](http://www.zerocracy.com/policy.html#36))

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @paulodamaso[/z](https://www.0crat.com/u/paulodamaso), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

sereshqua:
@0crat quality bad

0crat:
Quality is low, no payment, see [§31](http://www.zerocracy.com/policy.html#31)

PR #216, opened 2020-07-23
#215 - KeyLastPart introduced

### LINKED ISSUE #215: Last part of the key
Let's introduce class to obtain last part of the key (filename), this feature is required both in python and rpm-adapters.

olenagerasimova:
To close #215 I've added class to obtain last part of the key.

### DIFF
--- a/src/main/java/com/artipie/asto/ext/KeyLastPart.java
+++ b/src/main/java/com/artipie/asto/ext/KeyLastPart.java
@@ -49,7 +49,7 @@ public final class KeyLastPart {
      * @return Key last part as string
      */
     public String get() {
-        final String[] parts = this.origin.string().split("/");
+        final String[] parts = this.origin.string().replaceAll("/$", "").split("/");
         return parts[parts.length - 1];
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/artipie/asto/ext/KeyLastPart.java
```diff
@@ -0,0 +1,55 @@
...
+    public KeyLastPart(final Key origin) {
+        this.origin = origin;
+    }
+
+    /**
+     * Get last part of the key.
+     * @return Key last part as string
+     */
+    public String get() {
+        final String[] parts = this.origin.string().split("/");
```
g4s8, line 52: @olenagerasimova I think we need to remove trailing `/` chars before to avoid empty last part, e.g. `/foo/bar/` an `/foo/bar` should be equal for this class

olenagerasimova:
@g4s8 key starting with `/` is not valid, that's why I removed only ending `/`

0crat:
Job was finished in 0 hours, bonus for fast delivery is possible (see [§36](http://www.zerocracy.com/policy.html#36))

0crat:
Job audit: performer @g4s8[/z](https://www.0crat.com/u/g4s8)[/z](https://www.0crat.com/u/g4s8) didn't make CR comments

Issue #214, opened 2020-07-22
Locker for the storage


olenagerasimova:
Implement storage locker to add lock to storage being updated, implementation should
1. check if storage is already locked
2. add lock key
3. release lock key

The simplest implementation can be found in rpm-adapter, check `com.artipie.rpm.misc.StorageLock`.
Consider introducing functionality to wait while storage is released.

olenagerasimova:
@0crat assign @olegmoz

olegmoz:
@g4s8 please assign me

0crat:
@olegmoz[/z](https://www.0crat.com/u/olegmoz) this job was assigned to you 7days ago. It will be taken away from you soon, unless you close it, see [§8](http://www.zerocracy.com/policy.html#8). Read [this](http://www.yegor256.com/2014/04/13/no-obligations-principle.html) and [this](http://www.yegor256.com/2014/11/24/principles-of-bug-tracking.html), please.

0crat:
Job [#214](https://github.com/artipie/asto/issues/214) is not in the agenda of @olegmoz[/z](https://www.0crat.com/u/olegmoz), can't set title

PR #223, opened 2020-08-10
#214 - Added Lock and StorageLock with acquiring

olegmoz:
Part of #214 
Added `Lock` interface and `StorageLock` implementation with acquiring implemented.

### DIFF
--- a/src/main/java/com/artipie/asto/lock/StorageLock.java
+++ b/src/main/java/com/artipie/asto/lock/StorageLock.java
@@ -1,0 +1,125 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


src/test/java/com/artipie/asto/lock/StorageLockTest.java
```diff
@@ -0,0 +1,84 @@
...
+        new StorageLock(this.storage, this.target).acquire().join();
+        MatcherAssert.assertThat(
+            this.storage.list(new StorageLock.ProposalsKey(this.target))
+                .toCompletableFuture().join(),
+            new IsNot<>(new IsEmptyCollection<>())
+        );
+    }
+
+    @Test
+    void shouldAcquireSameLockTwice() {
```
olenagerasimova, line 66: @olegmoz this test method lacks assertion, could you please add it?

src/main/java/com/artipie/asto/lock/StorageLock.java
```diff
@@ -0,0 +1,125 @@
...
+    private final String uuid;
+
+    /**
+     * Ctor.
+     *
+     * @param storage Storage.
+     * @param target Target key.
+     */
+    public StorageLock(final Storage storage, final Key target) {
+        this(storage, target, UUID.randomUUID().toString());
```
olenagerasimova, line 63: @olegmoz such implementation does not seem suitable for rpm: `Rpm#batchUpdate(com.artipie.asto.Key)` starts update rpm repository by given key and I need to be sure that this repo (by this key) is not updated simultaneously. It's not possible with `StorageLock` implementation as it will simply add new lock on each update request. Cashing `UUID <-> key` map is not a solution in case of several artipie instances working with single storage.

olegmoz: @olenagerasimova `StorageLock` supposed to be created for each request and UUIDs have to be different for each request. When one lock is acquired (e.g. there is a value created from some UUID) next one will fail to acquire. This case is illustrated in test `shouldFailAcquireLockIfOtherProposalExists`.

Actually lock we have in RPM at the moment works the same way, see https://github.com/artipie/rpm-adapter/blob/master/src/main/java/com/artipie/rpm/misc/StorageLock.java#L83

olegmoz:
@olenagerasimova thanks for the review, corrected and answered

src/main/java/com/artipie/asto/lock/StorageLock.java
olenagerasimova: @olegmoz sorry, I've read the code not very carefully

olenagerasimova:
@olegmoz thanks, looks good to me

olegmoz:
> @olegmoz thanks, looks good to me

@olenagerasimova please approve the PR, so I can merge it

olenagerasimova:
@olegmoz sorry, done

0crat:
Job `gh:artipie/asto#223` is not assigned, can't get performer

Issue #225, opened 2020-08-10
Storage cache


g4s8:
Implement generic cache in `Storage` to be used by proxy repositories like Maven proxy or Files proxy

g4s8:
@rultor release, tag=`0.23.6`

PR #229, opened 2020-08-11
Initial storage cache implementation

g4s8:
#225 - initial storage cache implementation:
 - created primitive `StorageCache` implementation to improve it later (reads from remote to storage and returns storage after that)
 - created basic interfaces `Cache` and `CacheControl`
 - unit tests for `StorageCache`

### DIFF
--- a/src/main/java/com/artipie/asto/cache/Cache.java
+++ b/src/main/java/com/artipie/asto/cache/Cache.java
@@ -1,0 +1,56 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.artipie.asto.cache;
+
+import com.artipie.asto.Content;
+import com.artipie.asto.Key;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Supplier;
+
+/**
+ * Generic reactive cache which returns cached content by key of exist or loads from remote and
+ * cache if doesn't exit.
+ *
+ * @since 0.24
+ */
+public interface Cache {
+
+    /**
+     * No cache, just load remote resource.
+     */
+    Cache NOP = (key, remote, ctl) -> remote.get();
+
+    /**
+     * Try to load content from cache or fallback to remote publisher if cached key doesn't exist.
+     * When loading remote item, the cache may save its content to the cache storage.
+     * @param key Cached item key
+     * @param remote Remote source
+     * @param control Cache control
+     * @return Content for key
+     */
+    CompletionStage<? extends Content> load(
+        Key key, Supplier<? extends CompletionStage<? extends Content>> remote,
+        CacheControl control
+    );
+}
--- a/src/main/java/com/artipie/asto/cache/CacheControl.java
+++ b/src/main/java/com/artipie/asto/cache/CacheControl.java
@@ -1,0 +1,75 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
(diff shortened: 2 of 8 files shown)


src/main/java/com/artipie/asto/cache/StorageCache.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public CompletionStage<? extends Content> load(final Key key,
+        final Supplier<? extends CompletionStage<? extends Content>> remote,
+        final CacheControl control) {
+        final RxStorageWrapper rxsto = new RxStorageWrapper(this.storage);
+        return rxsto.exists(key)
+            .filter(exists -> exists)
+            .flatMapSingleElement(
+                exists -> SingleInterop.fromFuture(control.validate(key))
+                    .map(valid -> exists && valid)
```
olegmoz, line 62: @g4s8 this `.map(...)` call could be removed, because `exists` is always `true` at this point

src/main/java/com/artipie/asto/cache/StorageCache.java
```diff
@@ -0,0 +1,74 @@
...
+            )
+            .filter(valid -> valid)
+            .flatMapSingleElement(ignore -> rxsto.value(key))
+            .switchIfEmpty(
```
olegmoz, line 66: @g4s8 we should fallback to remote not only value is absent in cache or invalid, but also when reading from cache fails

src/main/java/com/artipie/asto/cache/StorageCache.java
```diff
@@ -0,0 +1,74 @@
...
+                SingleInterop.fromFuture(remote.get()).flatMapCompletable(
```
olegmoz, line 67: @g4s8 we should fallback to remote not only when the key is absent from cache or invalid, but also when reading from cache fails with an error. Ideally this should also happen when error occurs in the middle of reading content, but this is complicated to implement

src/main/java/com/artipie/asto/cache/StorageCache.java
```diff
@@ -0,0 +1,74 @@
...
+                    content -> rxsto.save(
+                        key, new Content.From(content.size(), content)
+                    )
```
olegmoz, line 70: @g4s8 here we are waiting for content to be saved into cache before giving away the content. This makes system using cache both slower and less resilient. Maybe it would be better to do the saving in async manner and return `remote.get()` right away?

src/main/java/com/artipie/asto/test/ContentIs.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public void describeTo(final Description description) {
+        description.appendText("has bytes ").appendValue(this.matcher);
+    }
+
+    @Override
+    public boolean matchesSafely(final Content item) {
+        final byte[] actual = new Concatenation(item).single()
+            .map(buf -> new Remaining(buf).bytes()).blockingGet();
+        return this.matcher.matches(actual);
```
olegmoz, line 80: @g4s8 there is nice `PublisherAs` class for reading bytes from `Content`, using it will make this method even less verbose

src/test/java/com/artipie/asto/cache/StorageCacheTest.java
```diff
@@ -0,0 +1,177 @@
...
+ * @checkstyle MagicNumberCheck (500 lines)
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+final class StorageCacheTest {
+
+    @Test
+    void loadsFromCache() throws Exception {
+        final Storage storage = new InMemoryStorage();
+        final Key key = new Key.From("key1");
```
olegmoz, line 61: @g4s8 we have a `Storage` and a `Key` in every test, I'd suggest to initialize them once in class fields

src/main/java/com/artipie/asto/cache/Cache.java
```diff
@@ -0,0 +1,56 @@
...
+     * When loading remote item, the cache may save its content to the cache storage.
+     * @param key Cached item key
+     * @param remote Remote source
+     * @param control Cache control
+     * @return Content for key
+     */
+    CompletionStage<? extends Content> load(
+        Key key, Supplier<? extends CompletionStage<? extends Content>> remote,
+        CacheControl control
+    );
```
olegmoz, line 55: @g4s8 design of this method seems a bit off: we pass a `Key` to the method and method always pass the same `Key` to `CacheControl` we have provided. Or there is scenario when we pass some other `Key` to the `CacheControl`? Cannot think of one. Maybe it would be better to specify `CacheControl` only once in constructor?

src/main/java/com/artipie/asto/cache/StorageCache.java
g4s8: @olegmoz yes, I just wanted to implement tests for this cache and improve it in another PR

src/main/java/com/artipie/asto/cache/Cache.java
g4s8: @olegmoz I created this `CacheControl` to respect [cache-control HTTP directives](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) sent by client: e.g. if we have some `Cache` instance for a `Slice`, and this slice receives two requests from the client: one request with `Cache-Control: max-age=<X>` header, and another with `Cache-Control: no-cache ` header, so in first case we need to check that cached item was cached `X` seconds at the latest, but for the seconds request we have to invalidate the cache. If we move it to constructor, then we need to instantiate cache object per request, but it seems not logically correct to create new cache for repository on request, WDYT?

g4s8:
@olegmoz thanks, fixed and answered, please check new commit

src/main/java/com/artipie/asto/cache/Cache.java
olegmoz: @g4s8 I see, but still do not see how `CacheControl` is a function of `Key`? Don't we always know value before invoking `load` method? Don't we always invoke `CacheControl` with the same `Key` we passed as first argument?

src/main/java/com/artipie/asto/cache/CacheControl.java
```diff
@@ -0,0 +1,76 @@
...
+        public CompletionStage<Boolean> validate(final Key item) {
+            return this.origin.validate(item);
+        }
+    }
+
+    /**
+     * Validate cached item: checks if cached value can be used or needs to be updated by fresh
+     * value.
+     * @param item Cached item
+     * @return True if cached item can be user, false if needs to be updated
```
olegmoz, line 73: @g4s8 it seems there is a little typo, should be `used` instead of `user`

src/main/java/com/artipie/asto/cache/Cache.java
g4s8: @olegmoz the idea of this `CacheControl` class was to delegate responsibility of validating cached items to separate objects and to create unlimited amount of implementations depends on client app needs. Before I thought about creating `enum` with possible cache-control directives, but with this approach the implementation will be located in all  `Cache` implementations.

src/main/java/com/artipie/asto/cache/CacheControl.java
g4s8: @olegmoz thanks, fixed

0crat:
Job `gh:artipie/asto#229` is not assigned, can't get performer

PR #230, opened 2020-08-13
#214 - Added RetryLock

### LINKED ISSUE #214: Locker for the storage

olegmoz:
Closes #214 
Added `RetryLock` that allows to make multiple acquire attempts on faulty origin lock

### DIFF
--- a/src/main/java/com/artipie/asto/lock/RetryLock.java
+++ b/src/main/java/com/artipie/asto/lock/RetryLock.java
@@ -1,0 +1,105 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 artipie.com
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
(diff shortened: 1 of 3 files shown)


src/main/java/com/artipie/asto/lock/RetryLock.java
```diff
@@ -0,0 +1,105 @@
...
+    public CompletionStage<Void> acquire() {
+        return this.registry.retry("lock-acquire").executeCompletionStage(
+            this.scheduler,
+            this.origin::acquire
+        );
+    }
+
+    @Override
+    public CompletionStage<Void> release() {
+        return this.origin.release();
```
g4s8, line 103: @olegmoz don't we need to retry `release` too? It seems `origin.release()` may fail the same way as `acquire()`

src/test/java/com/artipie/asto/lock/RetryLockTest.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+        ).acquire().toCompletableFuture().join();
+        MatcherAssert.assertThat(
+            attempt.get(),
+            new IsEqual<>(attempts)
+        );
+    }
+
+    @Test
+    @Timeout(10)
```
g4s8, line 100: @olegmoz can we make shorter interval test timeout? 10 seconds seems to be quite long

olegmoz:
@g4s8 thanks, added retries for `release` operation as well

0crat:
Job `gh:artipie/asto#230` is not assigned, can't get performer

Issue #233, opened 2020-08-17
Cache verification


g4s8:
Add additional parameter for cache to verify it's content. It'll be useful when the client knows hash-sums of cached data and want's to verify it before downloading.

Related to https://github.com/artipie/maven-adapter/issues/146

g4s8:
@rultor release, tag=`0.24`

0crat:
Job was finished in 42 hours, bonus for fast delivery is possible (see [§36](http://www.zerocracy.com/policy.html#36))

0crat:
There is an unrecoverable failure on my side. Please, submit it [here](https://github.com/zerocracy/farm/issues):

```
PID: 4@f4d970db-062a-4a00-a6aa-acaa32bbc384, thread: PQ-CT2E6TK9B
com.zerocracy.farm.ruled.RdRules[130] java.lang.IllegalStateException: http://datum.zerocracy.com/latest/rules/pm/no-lost-boosts.xsl says: The boost for job "gh:artipie/asto#233" exists in boosts.xml, but the job is absent in the WBS among 59 jobs.
```

[0.54.5](https://github.com/zerocracy/farm): CID: [fc2053ac-7924-4d0f-a533-8f29b25c0f24](https://www.0crat.com/footprint/CT2E6TK9B/fc2053ac-7924-4d0f-a533-8f29b25c0f24), Type: "Remove job from WBS"

PR #234, opened 2020-08-17
Digest cache verification

g4s8:
#233 - verify cache by message digest

### DIFF
--- a/src/main/java/com/artipie/asto/cache/CacheControl.java
+++ b/src/main/java/com/artipie/asto/cache/CacheControl.java
@@ -23,7 +23,7 @@
  */
 package com.artipie.asto.cache;
 
-import com.artipie.asto.Content;
+import com.artipie.asto.AsyncContent;
 import com.artipie.asto.Key;
 import hu.akarnokd.rxjava2.interop.SingleInterop;
 import io.reactivex.Observable;
@@ -31,13 +31,21 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
-import java.util.function.Supplier;
 
 /**
  * Cache control.
  * @since 0.24
  */
 public interface CacheControl {
+
+    /**
+     * Validate cached item: checks if cached value can be used or needs to be updated by fresh
+     * value.
+     * @param item Cached item
+     * @param content Content supplier
+     * @return True if cached item can be used, false if needs to be updated
+     */
+    CompletionStage<Boolean> validate(Key item, AsyncContent content);
 
     /**
      * Standard cache controls.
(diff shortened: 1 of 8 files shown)


src/main/java/com/artipie/asto/cache/CacheControl.java
```diff
@@ -61,16 +67,56 @@
...
     /**
      * Validate cached item: checks if cached value can be used or needs to be updated by fresh
      * value.
      * @param item Cached item
+     * @param content Content supplier
      * @return True if cached item can be used, false if needs to be updated
      */
-    CompletionStage<Boolean> validate(Key item);
+    CompletionStage<Boolean> validate(Key item,
+        Supplier<? extends CompletionStage<? extends Content>> content);
```
olegmoz, line 84: @g4s8 `Supplier<? extends CompletionStage<? extends Content>>` type looks to be a little bit too complex, what do you think about introducing an interface for this type?

src/main/java/com/artipie/asto/cache/CacheControl.java
```diff
@@ -61,16 +67,56 @@
...
+            this.items = items;
+        }
+
+        @Override
+        public CompletionStage<Boolean> validate(final Key key,
+            final Supplier<? extends CompletionStage<? extends Content>> content) {
+            return Observable.fromIterable(this.items)
+                .flatMapSingle(item -> SingleInterop.fromFuture(item.validate(key, content)))
+                .all(item -> item)
+                .to(SingleInterop.get());
```
olegmoz, line 119: @g4s8 it might be easier to compose `item.validate(...)` futures using `CompletableFuture.allOf()` without using reactive abstractions

src/main/java/com/artipie/asto/cache/CacheControl.java
```diff
@@ -61,16 +67,56 @@
...
      */
-    CompletionStage<Boolean> validate(Key item);
+    CompletionStage<Boolean> validate(Key item,
+        Supplier<? extends CompletionStage<? extends Content>> content);
+
+    /**
+     * All cache controls should validate the cache.
+     * @since 0.25
+     */
+    final class All implements CacheControl {
```
olegmoz, line 90: @g4s8 this class lacks test coverage

src/main/java/com/artipie/asto/cache/CacheControl.java
g4s8: @olegmoz I'm not sure it's possible, since `ompletableFuture.allOf()` reutrns `CompletableFuture<Void>` but I need `CopletableFuture<Boolean>` here

g4s8:
@olegmoz thanks, should be fixed in https://github.com/artipie/asto/pull/234/commits/fd3cf7a9a924cfbc859fdeda551e68c321eabd80

src/main/java/com/artipie/asto/cache/CacheControl.java
olegmoz: @g4s8 oh, that's right, missed that

0crat:
Job `gh:artipie/asto#234` is not assigned, can't get performer

PR #236, opened 2020-08-19
#231 - Using expiration time when acquiring StorageLock

### LINKED ISSUE #231: StorageLock expiration
When `StorageLock` is acquired by process and the process does not release it for whatever reason the lock cannot be obtained by any other process forever. Only manual deletion of lock from storage may solve the issue. The lock should expire automatically, after some timeout to allow timely release of the lock if some process didn't release it

olegmoz:
Closes #231 
Using expiration time when acquiring `StorageLock`. Acquiring fails only if not expired proposals exists.

### DIFF
--- a/src/main/java/com/artipie/asto/lock/storage/Proposals.java
+++ b/src/main/java/com/artipie/asto/lock/storage/Proposals.java
@@ -26,6 +26,7 @@ package com.artipie.asto.lock.storage;
 import com.artipie.asto.Content;
 import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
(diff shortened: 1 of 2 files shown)


olenagerasimova (approved):
@olegmoz thanks, looks good to merge

g4s8 (changes requested):
@olegmoz please check my comment

olegmoz:
@g4s8 thanks, it will be more correct to have single `now` for all expiration checks

g4s8 (approved):
@olegmoz thanks

0crat:
Job `gh:artipie/asto#236` is not assigned, can't get performer

PR #247, opened 2020-08-28
#246 - Added RxLock

### LINKED ISSUE #246: Reactive lock to work better with RxStorage
ASTO library provides `RxStorage` wrapper to use instead of `Storage` and it is used in adapters. When using with `Lock` it is needed to adapt this kind of storage manually. It would be nice to have such adapter ready in library

olegmoz:
Closes #246 
Added `RxLock`

### DIFF
--- a/src/main/java/com/artipie/asto/lock/RxLock.java
+++ b/src/main/java/com/artipie/asto/lock/RxLock.java
@@ -27,55 +27,41 @@ import hu.akarnokd.rxjava2.interop.CompletableInterop;
 import io.reactivex.Completable;
 
 /**
- * Asynchronous lock that might be successfully obtained by one thread only at a time.
+ * Reactive adapter for {@link Lock}.
  *
  * @since 0.27
  */
-public interface RxLock {
+public final class RxLock {
+
+    /**
+     * Origin.
+     */
+    private final Lock origin;
+
+    /**
+     * Ctor.
+     *
+     * @param origin Origin.
+     */
+    public RxLock(final Lock origin) {
+        this.origin = origin;
+    }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/artipie/asto/lock/RxLock.java
```diff
@@ -0,0 +1,81 @@
...
+
+import hu.akarnokd.rxjava2.interop.CompletableInterop;
+import io.reactivex.Completable;
+
+/**
+ * Asynchronous lock that might be successfully obtained by one thread only at a time.
+ *
+ * @since 0.27
+ */
+public interface RxLock {
```
g4s8, line 34: @olegmoz do we really need an interface here? This class seems to be a wrapper for `Lock` to use it for RX operations, similar to `BlockingStorage` for `Storage`. I'd not create new interfaces if we won't inject it as a dependency or pass as method parameter ever.

g4s8 (changes requested):
@olegmoz please check my comment

src/main/java/com/artipie/asto/lock/RxLock.java
olegmoz: @g4s8 yes, I think you are right. No point in making it an interface

g4s8:
@olegmoz please check CI build logs, some tests are failing

olegmoz:
@g4s8 that was a test crash due to timeout, now all the checks pass

g4s8 (approved):
@olegmoz thanks

0crat:
Job `gh:artipie/asto#247` is not assigned, can't get performer

Issue #237, opened 2020-08-24
Implement storage transactions using StorageLock


olegmoz:
Basic transaction functions may be implemented using `StorageLock` class to acquire lock for specified keys and so avoid concurrent access to locked data

0crat:
You @olegmoz[/z](https://www.0crat.com/u/olegmoz) need to have one of these roles in order to do "Request order start": `ARC`, `PO`; I'm sorry to say this, but at the moment you've got no roles in this project

0crat:
Job `gh:artipie/asto#237` is not assigned, can't get performer

olegmoz:
@rultor release, tag=`0.27`

PR #250, opened 2020-08-31
#237 - Replaced Storage.transaction with Storage.exclusively

olegmoz:
Part of #237 
Replaced `Storage.transaction` with `Storage.exclusively`


src/main/java/com/artipie/asto/LoggingStorage.java
```diff
@@ -138,13 +138,13 @@ public LoggingStorage(final Level level, final Storage storage) {
     }
 
     @Override
-    public CompletableFuture<Transaction> transaction(final List<Key> keys) {
-        return this.storage.transaction(keys).thenApply(
+    public CompletionStage<Void> exclusively(
```
g4s8, line 141: @olegmoz maybe we can use `<T>` instead of `<Void>` here with `Function<Storage, CompletionStage<T>>` to return some result from storage if needed?

src/main/java/com/artipie/asto/rx/RxStorage.java
```diff
@@ -97,10 +97,14 @@
...
+     * Runs operation exclusively for specified key.
      *
-     * @param keys The keys regarding which transaction is atomic
-     * @return Transaction
+     * @param key Key which is scope of operation.
+     * @param operation Operation to be performed exclusively.
+     * @return Completion or error signal.
      */
-    Single<RxTransaction> transaction(List<Key> keys);
+    Completable exclusively(
```
g4s8, line 106: @olegmoz if we agree to use `<T>` as `CompletionStage` paramter in `Storage` interface, then we could use `Single<T>` or `Maybe<T>` here

g4s8 (changes requested):
@olegmoz please check my comments

olegmoz:
@g4s8 thanks for suggestion, added return value for the operation

g4s8 (approved):
@olegmoz thanks

PR #264, opened 2020-09-04
#260 - Handle concurrently deleted proposals in StorageLock

### LINKED ISSUE #260: StorageLock fails acquire because of deleted proposals
`StorageLock` reads proposals list and tries to read their content for expiration time after that. However proposals might be missing already and it causes `acquire` method to fail without good reason. This case should could be handled by additional check for existance of proposal value in `Storage` if `value` method has failed.

olegmoz:
Closes #260 
Handle concurrently deleted proposals in `StorageLock`


olenagerasimova (approved):
@olegmoz thanks, looks good to me

0crat:
Job `gh:artipie/asto#264` is not assigned, can't get performer

Issue #171, opened 2020-05-13
File storage is too slow


g4s8:
Copying 1Gb file via `FileStorage` takes about 10-13 seconds, while copying the same file from and to same disks and filesystems takes less than 1 second. Let's fix it.

0crat:
The job [#171](https://github.com/artipie/asto/issues/171) assigned to @g4s8[/z](https://www.0crat.com/u/g4s8), here is [why](https://www.0crat.com/footprint/CT2E6TK9B/3be7bde9-fe9f-4a1c-bc48-6e371c8b4107); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

g4s8:
@rultor release, tag=`0.20`

g4s8:
Fixed in `0.20`

0crat:
Job `gh:artipie/asto#171` is not assigned, can't get performer

PR #280, opened 2020-09-21
Revert signature changes made with BlockingStorage

### LINKED ISSUE #281: Back to the old BlockingStorage
BlockingStorage was initially designed for simplicity, in #171 method signatures have been modified and now you need to explicitly handle the exceptions, which might look awful in a tiny lambda body.

Let's roll back the changes made.

Sammers21:
Closes #281.

### DIFF
--- a/src/main/java/com/artipie/asto/blocking/BlockingStorage.java
+++ b/src/main/java/com/artipie/asto/blocking/BlockingStorage.java
@@ -26,12 +26,9 @@ package com.artipie.asto.blocking;
 import com.artipie.asto.Concatenation;
 import com.artipie.asto.Content;
 import com.artipie.asto.Key;
-import com.artipie.asto.OneTimePublisher;
 import com.artipie.asto.Remaining;
 import com.artipie.asto.Storage;
-import com.google.common.util.concurrent.UncheckedExecutionException;
 import java.util.Collection;
-import java.util.concurrent.ExecutionException;
 
 /**
  * More primitive and easy to use wrapper to use {@code Storage}.
@@ -59,14 +56,9 @@ public class BlockingStorage {
      *
      * @param key The key (file name)
      * @return TRUE if exists, FALSE otherwise
-     * @throws InterruptedException If thread was interrupted
      */
-    public boolean exists(final Key key) throws InterruptedException {
-        try {
-            return this.storage.exists(key).get();
-        } catch (final ExecutionException err) {
-            throw new UncheckedExecutionException(err);
-        }
+    public boolean exists(final Key key) {
+        return this.storage.exists(key).join();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


olegmoz:
@Sammers21 please explain reason for these changes. I do not see how it is related to #171 and word "horrible" is not really informative

Sammers21:
@olegmoz, #171 is the ticket number where the changes were introduced. Reverted with the same ticket number.

Sammers21:
@olegmoz, description has been updated

olegmoz:
> @Sammers21 please explain reason for these changes. I do not see how it is related to #171 and word "horrible" is not really informative

@Sammers21 it is sad that changes were made without proper explanation. Let's avoid this kind of irresponsible behavior and write down motivation for the changes

Sammers21:
> Let's avoid this kind of irresponsible behavior and write down motivation for the changes

@olegmoz, description has been updated

olegmoz:
@Sammers21 normally a problem that requires to be solved is described in an issue and PR description is about the way problem was solved. Also PR title is still non-informative and judgemental, I'd suggest to change it and avoid such wording in future

Sammers21:
@olegmoz, better now?

olegmoz:
@Sammers21 thanks, now the reason for is described. Though I do not understand what word `sign` is there for

src/main/java/com/artipie/asto/blocking/BlockingStorage.java
```diff
@@ -43,30 +42,25 @@
     /**
      * Wrapped storage.
      */
-    private final Storage storage;
+    private final RxStorage storage;
```
olegmoz, line 45: @Sammers21 could you please explain why we need to use `RxStorage` here? It will be way more simple to just keep using normal `Storage`, but to invoke `.join()` on it's results instead of `.get()`

olegmoz (changes requested):
@Sammers21 thanks, please see one comment

src/main/java/com/artipie/asto/blocking/BlockingStorage.java
Sammers21: @olegmoz, because `join` is a bad name for the operation. `CompletableFuture` in general is a badly-designed primitive for reactive programming and should be avoided where possible.

olegmoz: @Sammers21 sorry, can't agree with you on `join` name being bad. It resembles "join thread" I believe. And `CompletableFuture` is not reactive programming primitive. Though it does not matter, neither `Storage` nor `BlockingStorage` are reactive and there is no sense to put something "reactive" in-between: it's unnecessary complication of code and waste of computation resources.

olegmoz (changes requested):
@Sammers21 please simplify `BlockingStorage` so it invokes `Storage` directly without wrapping it into `RxStorage`. Also it would be nice to remove unnecessary checked exception handling in all cases of `BlockingStorage` usage

olegmoz (changes requested):
@Sammers21 thanks, please remove emoji from commit message and fix the build - it is failing for some reason

Sammers21:
> @Sammers21 thanks, please remove emoji from commit message and fix the build - it is failing for some reason

why should I remove the emoji?

Sammers21:
@olegmoz, can we merge?

olegmoz (approved):
@Sammers21 thanks, merging

0crat:
Job `gh:artipie/asto#280` is not assigned, can't get performer

PR #283, opened 2020-09-23
Atomicity check

### LINKED ISSUE #282: Atomicity check test
There is no atomicity check test in the test suit. We should have one.

Sammers21:
Closes #282

### DIFF
--- a/src/test/java/com/artipie/asto/StorageSaveAndLoadTest.java
+++ b/src/test/java/com/artipie/asto/StorageSaveAndLoadTest.java
@@ -41,8 +41,10 @@ import org.junit.jupiter.api.extension.ExtendWith;
 /**
  * Tests for {@link Storage#save(Key, Content)} and {@link Storage#value(Key)}.
  *
+ * @checkstyle IllegalCatchCheck (500 lines)
  * @since 0.14
  */
+@SuppressWarnings("PMD.AvoidCatchingGenericException")
 @ExtendWith(StorageExtension.class)
 public final class StorageSaveAndLoadTest {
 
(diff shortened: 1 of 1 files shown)


src/test/java/com/artipie/asto/StorageSaveAndLoadTest.java
```diff
@@ -80,6 +83,39 @@ void shouldSaveFromMultipleBuffers(final Storage storage) throws Exception {
...
+                key,
+                new Content.OneTime(
+                    new Content.From(
+                        Flowable.concat(
+                            Flowable.just(ByteBuffer.wrap(new byte[]{1})),
+                            Flowable.error(new IllegalStateException())
+                        )
+                    )
+                )
+            ).get();
```
olegmoz, line 109: @Sammers21 we can use `.join()` here and avoid `throws` in method signature

src/test/java/com/artipie/asto/StorageSaveAndLoadTest.java
```diff
@@ -80,6 +83,39 @@ void shouldSaveFromMultipleBuffers(final Storage storage) throws Exception {
         );
     }
 
+    @SuppressWarnings("PMD.EmptyCatchBlock")
+    @TestTemplate
+    @Timeout(1)
+    void saveIsAtomic(final Storage storage) throws ExecutionException, InterruptedException {
```
olegmoz, line 89: @Sammers21 this test is very similar to `shouldNotSavePartial` test we already have. The only difference is that we check situation with previously existing value. I'd suggest to rename this method name `shouldNotOverwriteWithPartial` so similar tests will have similar names

olegmoz (changes requested):
@Sammers21 thanks, please see comments

Sammers21:
@olegmoz, review the PR one more time, please.

olegmoz (approved):
@Sammers21 thanks, merging

0crat:
Job `gh:artipie/asto#283` is not assigned, can't get performer

PR #285, opened 2020-10-12
#284 - cache impl to load from remote

### LINKED ISSUE #284: `Cache` implementation
Let's add implementation of `Cache` to always load content from remote and cache it to storage. Content should be obtained from storage cache only if remote failed to return item.

olenagerasimova:
Closes #284 
Added implementation of `Cache` to always load content from remote and add to storage and get from cache only if remote failed to return content.

### DIFF
--- a/src/main/java/com/artipie/asto/AsyncContent.java
+++ b/src/main/java/com/artipie/asto/AsyncContent.java
@@ -43,4 +43,31 @@ public interface AsyncContent extends Supplier<CompletionStage<? extends Content
 
     @Override
     CompletionStage<? extends Content> get();
+
+    /**
+     * Failed async content.
+     * @since 0.3
+     */
+    final class Failed implements AsyncContent {
+
+        /**
+         * Failure cause.
+         */
+        private final Throwable reason;
+
+        /**
+         * Ctor.
+         * @param reason Failure cause
+         */
+        public Failed(final Throwable reason) {
+            this.reason = reason;
+        }
+
+        @Override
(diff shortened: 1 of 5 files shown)


src/main/java/com/artipie/asto/AsyncContent.java
```diff
@@ -43,4 +43,31 @@
 
     @Override
     CompletionStage<? extends Content> get();
+
+    /**
+     * Failed async content.
+     * @since 0.3
```
olegmoz, line 49: @olenagerasimova `@since` version is off, it should be `0.30`. Same in other classes added in this PR

src/main/java/com/artipie/asto/cache/FromStorageCache.java
```diff
@@ -33,10 +33,12 @@
 import java.util.concurrent.CompletionStage;
 
 /**
- * Storage cache implementation.
+ * Cache implementation that trays to obtain items from storage cache,
```
olegmoz, line 36: @olenagerasimova there is typo, "trays" instead of "tries"

olegmoz (changes requested):
@olenagerasimova thanks, please see two minor comments

olenagerasimova:
@olegmoz thanks, corrected

olegmoz (approved):
@olenagerasimova thanks, merging

0crat:
Job `gh:artipie/asto#285` is not assigned, can't get performer

Issue #286, opened 2020-10-29
Cache should return Optional of Content


olenagerasimova:
Let's change signature of `Cache#load` method to return `CompletionStage` of Optional with Content as it's possible there will not be any content to return in cases we do not have requested data in cache and remote return any unsuccessful response.

0crat:
Job `gh:artipie/asto#286` is not assigned, can't get performer

olenagerasimova:
@rultor release, tag=`0.32`

PR #291, opened 2020-11-05
#286 - changed Cache to use Remote

olenagerasimova:
Part of #286 
Changed `Cache` signature to use `Remote`, corrected implementations and added `Remote.Failed`.

### DIFF
--- a/src/main/java/com/artipie/asto/cache/FromRemoteCache.java
+++ b/src/main/java/com/artipie/asto/cache/FromRemoteCache.java
@@ -59,12 +59,18 @@ public final class FromRemoteCache implements Cache {
                 final CompletionStage<Optional<? extends Content>> res;
                 if (throwable == null && content.isPresent()) {
                     res = this.storage.save(
-                        key,  new Content.From(content.get().size(), content.get())
+                        key, new Content.From(content.get().size(), content.get())
                     ).thenCompose(nothing -> this.storage.value(key))
                         .thenApply(Optional::of);
                 } else {
+                    final Throwable error;
+                    if (throwable == null) {
+                        error = new IllegalStateException("Failed to load content from remote");
+                    } else {
+                        error = throwable;
+                    }
                     res = new FromStorageCache(this.storage)
-                        .load(key, new Remote.Failed(throwable), control);
+                        .load(key, new Remote.Failed(error), control);
                 }
                 return res;
             }
--- a/src/main/java/com/artipie/asto/cache/FromRemoteCache.java
+++ b/src/main/java/com/artipie/asto/cache/FromRemoteCache.java
@@ -23,10 +23,10 @@
  */
 package com.artipie.asto.cache;
 
-import com.artipie.asto.AsyncContent;
 import com.artipie.asto.Content;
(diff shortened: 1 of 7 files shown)


src/main/java/com/artipie/asto/cache/Remote.java
```diff
@@ -26,14 +26,15 @@
...
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Supplier;
 
 /**
  * Async {@link java.util.function.Supplier} of {@link java.util.concurrent.CompletionStage}
  * with {@link Optional} of {@link Content}. It's a {@link FunctionalInterface}.
  *
- * @since 0.4
+ * @since 0.40
```
olegmoz, line 37: @olenagerasimova latest version is `0.31`, so I guess proper `@since` tag version is `0.32`

src/main/java/com/artipie/asto/cache/FromRemoteCache.java
```diff
@@ -51,18 +51,20 @@ public FromRemoteCache(final Storage storage) {
...
+                final CompletionStage<Optional<? extends Content>> res;
+                if (throwable == null && content.isPresent()) {
+                    res = this.storage.save(
+                        key,  new Content.From(content.get().size(), content.get())
+                    ).thenCompose(nothing -> this.storage.value(key))
+                        .thenApply(Optional::of);
                 } else {
                     res = new FromStorageCache(this.storage)
-                        .load(key, new AsyncContent.Failed(throwable), control);
+                        .load(key, new Remote.Failed(throwable), control);
```
olegmoz, line 67: @olenagerasimova we are going to this code branch when `content.isPresent() == false` and then `throwable` is `null`. Case when `throwable` is `null` and content is missing requires special handling. Also this case is not covered with tests

olegmoz (changes requested):
@olenagerasimova thanks, please see two comments

src/main/java/com/artipie/asto/cache/Remote.java
olenagerasimova: @olegmoz hm, in my experience `@sinse` tag usually contains next major version, in our case it's `0.40`

olegmoz: @olegmoz common version format is `<major>.<minor>.<patch>`. Next major version for us is `1.0` and next minor is `0.32`. We are releasing minor versions every time, but they happen to have breaking changes. I see how `@since` tag in this case might be `1.0` or `0.32`, but `0.40` just does not make sense to me, sorry

olenagerasimova:
@olegmoz thanks, corrected and aswered

src/main/java/com/artipie/asto/cache/Remote.java
olenagerasimova: @olegmoz ok, agreed, corrected since tags

olegmoz (approved):
@olenagerasimova thanks, merging

0crat:
Job `gh:artipie/asto#291` is not assigned, can't get performer

PR #298, opened 2021-01-27
Check ROOT key for save and load

### LINKED ISSUE #201: SubStorage fails with ROOT key
`SubStorage` from `ROOT` key fails to save content with `IllegalStateException`, I do not think this is correct behavior: `SubStorage` should process such case properly, as it's user I do not want to care whether my prefix is `ROOT` or not.

g4s8:
Closes #201 - added check for `ROOT` key in `save` and `value` storage
methods:
 - added tests to find all problem methods in `StorageSaveAndLoadTest`
 - fixed all `Storage` implementation to fail on `ROOT` key

### DIFF
--- a/src/main/java/com/artipie/asto/SubStorage.java
+++ b/src/main/java/com/artipie/asto/SubStorage.java
@@ -24,6 +24,7 @@
 package com.artipie.asto;
 
 import com.artipie.asto.lock.storage.StorageLock;
+import java.io.IOException;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


olenagerasimova:
@g4s8 seems good to merge, please fix the build

g4s8:
@olenagerasimova fixed, please check now

olenagerasimova (approved):
@g4s8 thanks

0crat:
Job `gh:artipie/asto#298` is not assigned, can't get performer

Issue #299, opened 2021-02-06
Tests from FromStorageCacheTest hang and never finish


andreoss:
The tests hang and never finish.
https://github.com/artipie/asto/blob/master/src/test/java/com/artipie/asto/cache/FromStorageCacheTest.java#L156-L163

Reproducible with these JDKs.
```
* $JDK_11/bin/java -version
openjdk version "11.0.9" 2020-10-20
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9+11)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.9+11, mixed mode)
* $JDK_15/bin/java -version
openjdk version "15.0.1" 2020-10-20
OpenJDK Runtime Environment AdoptOpenJDK (build 15.0.1+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 15.0.1+9, mixed mode, sharing)
issue-299 src/artipie/asto

```

g4s8:
@andreoss thanks for reporting. It looks strange: I can't reproduce it locally, and CI passes fine: https://github.com/artipie/asto/actions?query=workflow%3A%22Maven+Build%22
What is your environment?

andreoss:
@g4s8 Reproducible like that
```
$ src/artipie/asto
* docker run -it --rm -v $HOME/.m2:/root/.m2 -v "$(pwd)":"/tmp/$(pwd)"   -w "/tmp/$PWD"  maven  mvn clean install -Pqulice
```

g4s8:
@andreoss I tried to run the same command three times, and it finished successfully:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  34.542 s
[INFO] Finished at: 2021-02-08T12:13:27Z
[INFO] ------------------------------------------------------------------------
```

Can you please check you're using latest version in master:
```
git rev-parse HEAD
bbd2af4e584aee0f240a663f6e1063ee3a1a82f3
```

andreoss:
@g4s8 
It depends on the number of available CPUs

Dead-locks: 
```
* docker run  --cpus=8 -it --rm        -v $HOME/.m2:/root/.m2 -v "$(pwd)":"/tmp/$(pwd)"        -w "/tmp/$PWD"        maven        mvn clean test -Dtest='*FromStorageCacheTest'
```
Passes: 
```
[...]

PR #300, opened 2021-02-06
#299: Prevent dead lock in tests

andreoss:
Per #299:
- Prevent dead lock
- Replace `CountDownLatch` with `Semaphore`

### DIFF
--- a/src/test/java/com/artipie/asto/cache/FromStorageCacheTest.java
+++ b/src/test/java/com/artipie/asto/cache/FromStorageCacheTest.java
@@ -137,19 +137,22 @@ final class FromStorageCacheTest {
         final FromStorageCache cache = new FromStorageCache(this.storage);
         final Key key = new Key.From("key4");
         final int count = 100;
-        final CountDownLatch latch = new CountDownLatch(Runtime.getRuntime().availableProcessors() - 1);
+        final CountDownLatch latch = new CountDownLatch(
+            Runtime.getRuntime().availableProcessors() - 1
+        );
         final byte[] data = "data".getBytes();
         final Remote remote =
             () -> CompletableFuture
-                .runAsync(() -> {
-                    latch.countDown();
-                    try {
-                        latch.await();
-                    } catch (final InterruptedException ex) {
-                        Thread.currentThread().interrupt();
-                        throw new IllegalStateException(ex);
-                    }
-                })
+                .runAsync(
+                    () -> {
+                        latch.countDown();
+                        try {
+                            latch.await();
+                        } catch (final InterruptedException ex) {
+                            Thread.currentThread().interrupt();
+                            throw new IllegalStateException(ex);
+                        }
+                    })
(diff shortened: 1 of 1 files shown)


0crat:
@andreoss[/z](https://www.0crat.com/u/andreoss) this pull request is too small, just 2 lines changed (less than 10), there will be no formal code review, see [§53](http://www.zerocracy.com/policy.html#53) and [§28](http://www.zerocracy.com/policy.html#28); in the future, try to make sure your pull requests are not too small; @g4s8[/z](https://www.0crat.com/u/g4s8) please review this and merge or reject

genryxy (changes requested):
@andreoss thanks, the main idea of this test cases is that it is possible to process multiple requests simultaneously. We have many threads blocking on `await()` that would all start simultaneously when the `countDown` reached zero. But `Semaphore` only limits the number of threads that possible to enter the critical section. Therefore I thing it is wrong to replace `CountDownLatch` with `Semaphore` in this case.

andreoss:
@genryxy 
Thanks for explanation. 
It seemed that the idea here was to make `Remote` operate on a slower rate or emulate some kind of throttling.
This test will deadlock if it exhausts the standard thread pool (i.e the latch value is greater than available processors). 
Could be an issue only with OpenJDK or Linux, as it's what I use.

genryxy:
@andreoss got it. Probably to use `Runtime.getRuntime().availableProcessors()` to get available number. Or at least add `todo` and disable test.
@g4s8 please take a look

genryxy (approved):
@andreoss thanks, looks good to me
@g4s8 please check/merge

src/test/java/com/artipie/asto/cache/FromStorageCacheTest.java
```diff
@@ -137,22 +137,26 @@ void processMultipleRequestsSimultaneously() throws Exception {
         final FromStorageCache cache = new FromStorageCache(this.storage);
         final Key key = new Key.From("key4");
         final int count = 100;
-        final CountDownLatch latch = new CountDownLatch(10);
+        final CountDownLatch latch = new CountDownLatch(
+            Runtime.getRuntime().availableProcessors() - 1
```
g4s8, line 141: @andreoss what if any runtime has only one CPU, I think some CI isntances may have 1 CPU per builder?

g4s8 (changes requested):
@andreoss please check my comment

src/test/java/com/artipie/asto/cache/FromStorageCacheTest.java
andreoss: @g4s8 The latch will be no-op than. The test relys on common fork pool as I understand, and won't make much sense on a single-threaded machine

g4s8 (approved):
@andreoss got it, thanks

0crat:
Job `gh:artipie/asto#300` is not assigned, can't get performer

Issue #196, opened 2020-06-05
S3Storage.java:67-71: Too many methods in `S3Storage`...


0pdd:
The puzzle `193-ee41fbc9` from #193 has to be resolved: 

https://github.com/artipie/asto/blob/23201287aefb4b520aa2a9f70f8667014f750b05/src/main/java/com/artipie/asto/s3/S3Storage.java#L67-L71

The puzzle was created by @olegmoz on 04-Jun-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8) please, pay attention to this issue

victornoel:
@g4s8 I think you assigned to me instead of 0crat :)

andreoss:
@0crat await #299

andreoss:
@0crat wait for #299

0crat:
> @0crat await #299  ([here](https://github.com/artipie/asto/issues/196#issuecomment-774536493))

@andreoss Can't understand "await", try one of these:
  - `assign` Assign a performer to this issue
  - `boost` Set a boost for the job
  - `continue` Remove a job's impediment
  - `elect` Trigger election for ticket
  - `hello` Just say hello
  - `in` Register this issue in scope (WBS)
  - `out` Close the order and remove this job from scope
  - `pay` Pay a user some extra cash
  - `quality` Review a task
  - `resign` Resign from this issue
  - `status` Check the status of the job
  - `version` Print current version of the bot
  - `wait` Register an impediment for a job

0pdd:
The puzzle `193-ee41fbc9` has disappeared from the source code, that's why I closed this issue.

0crat:
Job [#196](https://github.com/artipie/asto/issues/196) is not in the agenda of @andreoss[/z](https://www.0crat.com/u/andreoss), can't set title

PR #301, opened 2021-02-09
#196: Extract objects from S3Storage

andreoss:
Per #196

- Replace static methods with objects decorating `Content`
- Extract handler of Exceptions into a class


src/main/java/com/artipie/asto/s3/S3Storage.java
```diff
@@ -386,69 +280,49 @@ public S3Storage(final S3AsyncClient client, final String bucket, final boolean
...
 
     /**
-     * Uploads content using multipart upload.
+     * Save multipart.
      *
-     * @param key Object key.
-     * @param content Object content to be uploaded.
-     * @return Completion stage which is completed when upload is completed or aborted.
+     * @param updated The estimated content.
+     * @param key The key.
```
genryxy, line 286: @andreoss I think it would be better to get a more detailed description of params

src/test/java/com/artipie/asto/StorageMoveTest.java
```diff
@@ -37,10 +37,10 @@
  * @since 0.14
  */
 @ExtendWith(StorageExtension.class)
+@Timeout(2)
 public final class StorageMoveTest {
```
genryxy, line 41: @andreoss why is it necessary to increase timeout here from one second to two?

src/test/java/com/artipie/asto/s3/ContentOfFutureTest.java
```diff
@@ -0,0 +1,72 @@
...
+        );
+        Assertions.assertArrayEquals(
+            Flowable.fromPublisher(content)
+                .singleElement()
+                .to(MaybeInterop.get())
+                .toCompletableFuture()
+                .join()
+                .array(),
+            data
+        );
```
genryxy, line 64: @andreoss please use `MatcherAssert.assertThat()`. Also it would be better to replace `Flowable.fromPublisher...` with `PublisherAs`

src/test/java/com/artipie/asto/s3/ContentOfFutureTest.java
```diff
@@ -0,0 +1,72 @@
...
+        MatcherAssert.assertThat(
+            content.size(), new IsEqual<>(Optional.of((long) data.length))
+        );
```
genryxy, line 67: @andreoss give description for assertions if you have more than one matchers in test method

src/test/java/com/artipie/asto/s3/InternalExceptionHandleTest.java
```diff
@@ -0,0 +1,61 @@
...
+                        IllegalStateException::new
+                    )
+                )
+                    .thenCompose(Function.identity())
+                    .toCompletableFuture()
+                    ::get
+            ),
+            Matchers.hasProperty("cause", Matchers.isA(IllegalStateException.class))
+        );
+    }
```
genryxy, line 60: @andreoss nested assertions don't look good.  You can capture the exception in ` AtomicReference` and after that make required checks

genryxy (changes requested):
@andreoss thanks, please see comments

src/test/java/com/artipie/asto/StorageMoveTest.java
andreoss: @genryxy It seems that this test is flakes, sometimes it fails on GHA.

src/main/java/com/artipie/asto/s3/ContentOfFuture.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ExecutionException;
+import org.reactivestreams.Subscriber;
+
+/**
+ * Content of a future.
+ *
+ * @since 1.0
+ */
```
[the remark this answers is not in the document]
andreoss: @genryxy I took that from pom.xml. What is the correct number? I'm always confused about that.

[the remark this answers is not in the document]
genryxy: @andreoss you can look [here](https://github.com/artipie/asto/releases) the latest available version.  And follow these [rules](https://semver.org/). Briefly, now you need to increase `minor` version (e.g. `0.34`)

src/test/java/com/artipie/asto/s3/InternalExceptionHandleTest.java
andreoss: @genryxy `assertThrows` returns caught Exception especially for the later assertion

genryxy: @andreoss okay, got it

src/main/java/com/artipie/asto/s3/ContentOfFuture.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Complete the future and handle checked exceptions.
+     *
+     * @return The content.
+     */
+    private Content complete() {
+        try {
+            return this.original.toCompletableFuture().get();
+        } catch (final InterruptedException ex) {
```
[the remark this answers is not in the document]
andreoss: @genryxy It would, but it's returned from another `Future`.

src/main/java/com/artipie/asto/s3/S3Storage.java
andreoss: @genryxy Done.

src/test/java/com/artipie/asto/s3/ContentOfFutureTest.java
andreoss: @genryxy Done.

src/test/java/com/artipie/asto/s3/ContentOfFutureTest.java
andreoss: @genryxy Done.
By the way, if you want  to require descriptions on assertion, you can do it with a static analyzer.
There's how cactoos does it.
https://github.com/yegor256/cactoos/blob/master/src/test/resources/forbidden-apis.txt#L4

genryxy:
@andreoss I do not see any pushed changes after previous review

src/main/java/com/artipie/asto/s3/ContentOfFuture.java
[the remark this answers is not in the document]
genryxy: @andreoss please redesign this class in some way to avoid using `get()` method in not test code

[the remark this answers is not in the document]
andreoss: @genryxy Any suggestions how? Can I leave a todo in order to continue working on it?
I don't see an issue here, because
-  `Content` seems to have a blocking API anyway 
- It is a part of a Future which is Returned from `Storage`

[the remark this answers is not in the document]
genryxy: @andreoss `Content` does not have a blocking API.  
I think this class may not implement `Content` interface and one of the method of this class could return `CompletionStage<Content>` because now you wrap estimated content into future:
```
result = CompletableFuture
                .completedFuture(new EstimatedContent(onetime))
                .thenCompose(updated -> this.put(key, updated));
```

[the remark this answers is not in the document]
andreoss: @genryxy Done.

src/main/java/com/artipie/asto/s3/ContentOfFuture.java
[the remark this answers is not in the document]
andreoss: @genryxy Done.

src/test/java/com/artipie/asto/RemainingTest.java
```diff
@@ -32,6 +32,9 @@
 /**
  * Test case for {@link Remaining}.
  * @since 1.0
+ * @todo #196:30m This since tag is incorrect.
+ *  Replace version here and in @{link KeyTest} with
+ *  correct number.
```
genryxy, line 37: @andreoss sorry, but I think it does not deserve to be a separate task. Probably to change the version change or leave it as it is without adding `@todo`

src/main/java/com/artipie/asto/s3/S3Storage.java
```diff
@@ -67,19 +57,14 @@
...
- * @todo #193:30min Too many methods in `S3Storage` class.
- *  `S3Storage` has too many methods.
- *  It could be refactored to move some of it's logic to other classes.
- *  In particular there three methods related to saving bytes
- *  (multipart or not, selecting proper method). This could be moved out of this class.
  * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ * @since 0.15
  */
 @SuppressWarnings("PMD.TooManyMethods")
 public final class S3Storage implements Storage {
```
genryxy, line 70: @andreoss is it possible to remove now this annotation?

genryxy (changes requested):
@andreoss thanks, please see two minor comments

genryxy (approved):
@andreoss thanks, looks good to me
@g4s8 please check/merge

src/main/java/com/artipie/asto/s3/S3Storage.java
andreoss: @genryxy Done

src/test/java/com/artipie/asto/RemainingTest.java
andreoss: @genryxy Removed

src/main/java/com/artipie/asto/s3/EstimatedContentCompliment.java
```diff
@@ -0,0 +1,164 @@
...
+                    }
+                    return size;
+                }
+            ).thenApply(
+                sizeOpt -> {
+                    final Flowable<ByteBuffer> data = Flowable.fromPublisher(
+                        new File(temp).content()
+                    ).doAfterTerminate(
+                        () -> Files.delete(temp)
+                    );
```
g4s8, line 140: @andreoss it seems we have the similar logic for file removing in two places. I'd remove this `afterTerminate` callback and remove the file in `handle`

src/main/java/com/artipie/asto/s3/EstimatedContentCompliment.java
```diff
@@ -0,0 +1,164 @@
...
+                }
+            )
+            .handle(
+                (value, throwable) -> {
+                    final Content result;
+                    if (throwable == null) {
+                        result = value;
+                    } else {
+                        try {
+                            Files.delete(temp);
```
g4s8, line 153: @andreoss here: we can remove tmp file in a single place without checking if error is not `null`

src/main/java/com/artipie/asto/s3/S3Storage.java
```diff
@@ -173,22 +157,16 @@ public S3Storage(final S3AsyncClient client, final String bucket, final boolean
...
-                    } else {
-                        future = this.uploadMultipart(key, updated);
-                    }
-                    return future;
-                }
-            );
+            result =
+                new EstimatedContentCompliment(onetime, S3Storage.MIN_MULTIPART)
+                    .estimate()
+                    .toCompletableFuture()
```
g4s8, line 163: @andreoss this call is not needed, `thenCompose` is a part of `CompletionStage` interface

src/main/java/com/artipie/asto/s3/S3Storage.java
```diff
@@ -173,22 +157,16 @@ public S3Storage(final S3AsyncClient client, final String bucket, final boolean
...
+                    .thenCompose(estimated -> this.putMultipart(estimated, key));
```
g4s8, line 164: @andreoss we have lost the part of logic here: previosly, S3 storage used simple `put` method for content with a size less than `S3Storage.MIN_MULTIPART` and `uploadMultipart` only for content greater or equal to this value. Now each upload request goes to `putMultipart` method, and it's not correct, since it's more expencive for small data uploads (it requires 3 HTTP requests: 1) create multipart upload, get upload ID 2) upload all parts to this ID 3) finish upload by ID; where simple upload is just a one HTTP call)

src/main/java/com/artipie/asto/s3/InternalExceptionHandle.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Ctor.
+     *
+     * @param future Original future.
+     * @param from Internal type of exception.
+     * @param convert Converter to a external type.
+     */
+    InternalExceptionHandle(
+        final CompletableFuture<T> future,
```
g4s8, line 64: @andreoss why do we need to depend on `future` here? `thenApply` method takes optional result and error and returns the result. In case of matched exception `this.from.isInstance(throwable.getCause())` we can throw converted exception from method body:
```java
    @Override
    public CompletionStage<T> apply(final T content, final Throwable throwable) {
        CompletionStage<T> result = this.future;
        if (
            throwable instanceof CompletionException
                && this.from.isInstance(throwable.getCause())
        ) {
            throw new FailedCompletionStage<>(
                this.convert.apply(throwable.getCause())
            );
        }
        return content;
    }
```

The problem with current code is that it adds implicit dependency on `future` and potenticaly may cause some side effects or unexpected behavior (e.g. if we cancel the future in this method or if we inject another future by the mistake)

src/main/java/com/artipie/asto/s3/S3Storage.java
```diff
@@ -215,15 +193,21 @@ public S3Storage(final S3AsyncClient client, final String bucket, final boolean
...
-                    .build()
-            ).thenApply(HeadObjectResponse::contentLength)
-        );
+        final CompletableFuture<Long> future = this.client.headObject(
+            HeadObjectRequest.builder()
+                .bucket(this.bucket)
+                .key(key.string())
+                .build()
+        ).thenApply(HeadObjectResponse::contentLength);
+        return future
```
g4s8, line 202: @andreoss there are not need to introduce a one-time variable, you can just call `handle` after `thenApply`

src/main/java/com/artipie/asto/s3/S3Storage.java
```diff
@@ -386,69 +282,49 @@ public S3Storage(final S3AsyncClient client, final String bucket, final boolean
...
-                    final Throwable cause = throwable.getCause();
-                    if (cause instanceof NoSuchKeyException) {
-                        result = new FailedCompletionStage<>(
-                            new ValueNotFoundException(key, cause)
-                        );
+    private CompletableFuture<Void> putMultipart(final Content updated, final Key key) {
+        final CompletableFuture<Void> future;
+        final Optional<Long> size = updated.size();
+        if (size.isPresent() && size.get() < S3Storage.MIN_MULTIPART) {
+            future = this.put(key, updated);
```
g4s8, line 295: @andreoss it's a bit confusing: the method is named `putMultipart` but here we're using non-multipart simple upload

g4s8 (changes requested):
@andreoss please check my comment. And please try to make pull requests a little smaller to avoid long reveiws (e.g. it could be splitted into 2 PRs: 1) introducing new classes and tests for the refactoring 2) apply the refactoring using new classes)

src/main/java/com/artipie/asto/s3/EstimatedContentCompliment.java
andreoss: @g4s8 Removing file in `handle` does not work. The file is removed too soon causing `Early EOF` exception

src/main/java/com/artipie/asto/s3/EstimatedContentCompliment.java
andreoss: @g4s8 I removed `handle` completely, and moved cleanup to `Flowable` hooks

src/main/java/com/artipie/asto/s3/S3Storage.java
andreoss: @g4s8 Done

src/main/java/com/artipie/asto/s3/S3Storage.java
andreoss: @g4s8 Done.

src/main/java/com/artipie/asto/s3/InternalExceptionHandle.java
andreoss: @g4s8 Done.

src/main/java/com/artipie/asto/s3/S3Storage.java
andreoss: @g4s8 Done

src/main/java/com/artipie/asto/s3/S3Storage.java
andreoss: @g4s8 Done

g4s8 (approved):
@andreoss thanks

0crat:
Job `gh:artipie/asto#301` is not assigned, can't get performer

PR #303, opened 2021-02-19
#302 - delete empty directories

### LINKED ISSUE #302: FileStorage: Delete empty directories
File storage should delete empty directories while deleting files

olenagerasimova:
Closes #302 
Delete empty directories along with files.

### DIFF
--- a/src/main/java/com/artipie/asto/fs/FileStorage.java
+++ b/src/main/java/com/artipie/asto/fs/FileStorage.java
@@ -205,11 +205,16 @@ public final class FileStorage implements Storage {
     public CompletableFuture<Void> delete(final Key key) {
         return CompletableFuture.runAsync(
             () -> {
-                try {
-                    Files.delete(this.path(key));
-                    this.deleteEmptyDirs(key.parent());
-                } catch (final IOException iex) {
-                    throw new UncheckedIOException(iex);
+                final Path path = this.path(key);
+                if (Files.exists(path) && !Files.isDirectory(path)) {
+                    try {
+                        Files.delete(path);
+                        this.deleteEmptyParts(key.parent());
+                    } catch (final IOException iex) {
+                        throw new UncheckedIOException(iex);
+                    }
+                } else {
+                    throw new ValueNotFoundException(key);
                 }
             },
             this.exec
@@ -294,16 +299,16 @@ public final class FileStorage implements Storage {
     }
 
     /**
-     * Removes empty directories.
+     * Removes empty key parts (directories).
      * @param key Key
      */
-    private void deleteEmptyDirs(final Optional<Key> key) {
+    private void deleteEmptyParts(final Optional<Key> key) {
         try {
             if (key.isPresent() && !key.get().string().isEmpty()
-                && this.path(key.get()).toFile().isDirectory()
+                && Files.isDirectory(this.path(key.get()))
                 && !Files.list(this.path(key.get())).findFirst().isPresent()) {
                 Files.delete(this.path(key.get()));
-                this.deleteEmptyDirs(key.get().parent());
+                this.deleteEmptyParts(key.get().parent());
             }
         } catch (final IOException err) {
             throw new UncheckedIOException(err);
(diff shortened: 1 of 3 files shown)


src/main/java/com/artipie/asto/fs/FileStorage.java
```diff
@@ -290,6 +293,23 @@ private Path path(final Key key) {
         return Paths.get(this.dir.toString(), key.string());
     }
 
+    /**
+     * Removes empty directories.
+     * @param key Key
+     */
+    private void deleteEmptyDirs(final Optional<Key> key) {
```
g4s8, line 300: @olenagerasimova I think we need to use `Path` parameter here, since the abstraction of file-system, path and directories is lower than storage and keys. When we're calling the method `deleteEmptyDirs`, where dir means directories, then we're using low-lever abstraction of the file-system and IO Java API

src/main/java/com/artipie/asto/fs/FileStorage.java
```diff
@@ -290,6 +293,23 @@ private Path path(final Key key) {
...
+        try {
+            if (key.isPresent() && !key.get().string().isEmpty()
+                && this.path(key.get()).toFile().isDirectory()
```
g4s8, line 303: @olenagerasimova you can check it with new Java API: `Files.isDirectory(path)`. `toFile()` method is needed for compatibility. Also, when using `File.isDirectory(path)` you don't need to check that `path` exist or not empty

src/main/java/com/artipie/asto/fs/FileStorage.java
```diff
@@ -290,6 +293,23 @@ private Path path(final Key key) {
...
+                && !Files.list(this.path(key.get())).findFirst().isPresent()) {
```
g4s8, line 304: @olenagerasimova we also need a check that the `path` we're going to remove is a child in hierarchy for `this.dir` storage root.

src/main/java/com/artipie/asto/fs/FileStorage.java
```diff
@@ -205,6 +207,7 @@ public FileStorage(final Path path, final ExecutorService exec) {
             () -> {
                 try {
                     Files.delete(this.path(key));
```
g4s8, line 209: @olenagerasimova I think we need to add a check that path of key we're going to remove here is not a directory for consistency with `exists` method

g4s8 (changes requested):
@olenagerasimova please check my comments

src/main/java/com/artipie/asto/fs/FileStorage.java
olenagerasimova: @g4s8 I'd prefer to change the naming, not the parameter. When there are no parents of an empty one of the key, that means, that we reached the storage root folder.

src/main/java/com/artipie/asto/fs/FileStorage.java
olenagerasimova: @g4s8 we are checking it with the help of `Key.perent()` method

olenagerasimova:
@g4s8 please check the changes and answers

src/main/java/com/artipie/asto/fs/FileStorage.java
```diff
@@ -290,6 +299,30 @@ private Path path(final Key key) {
...
     }
 
+    /**
+     * Removes empty key parts (directories).
+     * @param key Key
+     */
+    private void deleteEmptyParts(final Optional<Key> key) {
+        try {
+            if (key.isPresent() && !key.get().string().isEmpty()
+                && Files.isDirectory(this.path(key.get()))) {
```
g4s8, line 309: @olenagerasimova `this.path(key.get())` calls are used 3 times in this method, it could moved to a variable

g4s8 (changes requested):
@olenagerasimova just one minor comment

olenagerasimova:
@g4s8 corrected, please, take a look

g4s8 (approved):
@olenagerasimova thanks

0crat:
Job `gh:artipie/asto#303` is not assigned, can't get performer

PR #308, opened 2021-04-21
Do not throw `NoSuchFileException` while removing empty dirs in `FileStorage`

### LINKED ISSUE #307: `Delete` operation in `FileStorage`
It's possible to catch `NoSuchFileException` when `FileStorage.delete()` operation is performed in multithreads and all the files from the directory are deleted and it is required to remove the directory itself. Let's correct `FileStorage#deleteEmptyParts` not to throw `NoSuchFileException`.

olenagerasimova:
Closes #307 
Corrected `FileStorage#deleteEmptyParts()` not to throw `NoSuchFileException` if empty folder we are going to remove does not exist.


g4s8 (approved):
@olenagerasimova thanks

Issue #314, opened 2021-05-21
Create storage implementation for benchmarks


g4s8:
## Problem

Each adapter benchmark requires common logic for storage setup: it copies test data to storage, then run benchmark, then reset storage to initial state to run benchmark again. Using JMH, we can run multiple benchmarks simultaneously but in that case we need to construct and initialize a new storage for each test run, it may affect other benchmarks. Share one storage for all benchmarks is also not an option, since each run can write/delete/move data in storage.

## Proposal

So I'm proposing this solution: make a new copy-on-write storage implementation with a base shared read-only storage as a back-end, local isolated storage for write operations and a hash-table for deleted flags. So each JMH benchmark iteration will create and initialize a base storage with test data packages as a base read-only storage, and each benchmark run will create a new isolated local storage with base storage back-end for benchmark and a fresh table for delete-flags.
 - Each **read** operation first checks the deleted flags table and if flag exist, then fails with not-found exception, if deleted flag doesn't exist, then it tries to read from local storage, if local storage doesn't have such key, then go to shared back-end
 - Each **write** operation goes to local storage and rest deleted flag if exist
 - Each **delete** operation just put deleted flag to table
 - Each **move** operation tries to move data in local storage. If local key not found, then it copies data from shared storage to local first

Using `InMemory` storage as a back-end, we can optimize move operations by using data byte-array references for copying: since `InMemory` implementation doesn't perform in-place modification in data, we can safely copy byte-array reference from back-end to local storage on move.

### Architecture

```
+-----------+
|  Storage  |
+-----------+
      |
      V
+----------------+
|  BenchStorage  |
+----------------+
| backEnd        +------> InMemoryStorage
| local          +------> Map<Key, byte[]>
| deleted        +------> Set<Key>
+----------------+
```

g4s8:
@artipie/members what do you think?

genryxy:
@g4s8 I'd suggest to specify that it would be better to use `ConcurrentHashSet` from guava instead of `HashSet` to avoid synchronizing for each operation as we decided in Helm. 
And the same for `ConcurrentHashMap` instead of `HashMap`. Or `NavigableMap` if method `list` should be implemented.
Also I have concern about **move** operation because it seems there is no way to get reference on byte array for existed implementation of `InMemoryStorage`.

genryxy:
@g4s8 could you please release a new version

PR #315, opened 2021-05-24
Add implementation of  `save` and `value` for benchmark storage

genryxy:
Part of #314 
Added implementation of `save` and `value` for benchmark storage. 
Later add more tests for this implementation and implement other methods.


src/main/java/com/artipie/asto/Key.java
```diff
@@ -18,7 +19,7 @@
  *
  * @since 0.6
  */
-public interface Key {
+public interface Key extends Comparable<Key> {
```
g4s8, line 22: I'd suggest add `Comparator<? super Key>` interface static field instead of extending `Comparable<Key>` - this way you don't need to implement `compareTo` in each implementation and avoid confusing behavior of default interface implementation. It's a better strategy for comparing different interface implementation; where implementing `Comparable` is a good way for comparing `final` classes.

src/main/java/com/artipie/asto/bench/BenchmarkStorage.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Ctor.
+     * @param backend Backend storage
+     * @param local Content of local storage
+     */
+    public BenchmarkStorage(final InMemoryStorage backend, final Map<Key, byte[]> local) {
+        this.backend = backend;
+        this.local = new ConcurrentSkipListMap<>(local);
+        this.deleted = ConcurrentHashMap.newKeySet();
```
g4s8, line 67: Why not `new ConcurrentSkipListSet<>`?

src/main/java/com/artipie/asto/bench/BenchmarkStorage.java
```diff
@@ -0,0 +1,144 @@
...
+        throw new NotImplementedError("Not implemented yet");
+    }
+
+    @Override
+    public CompletableFuture<Content> value(final Key key) {
+        return CompletableFuture.runAsync(
+            () -> {
+                if (this.deleted.contains(key)) {
+                    throw new ValueNotFoundException(key);
+                }
```
g4s8, line 114: We can perform this check synchronously without submitting a new task in pool, set check for key is really fast. It may decrease benchmark's baseline.

src/main/java/com/artipie/asto/bench/BenchmarkStorage.java
```diff
@@ -0,0 +1,144 @@
...
+                final CompletableFuture<Content> res;
+                if (lcl == null) {
+                    res = this.backend.value(key);
+                } else {
+                    res = CompletableFuture.completedFuture(
+                        new Content.OneTime(new Content.From(lcl))
+                    );
+                }
+                return res;
+            }
```
g4s8, line 128: And same here: we can access byte-array reference from map by key almost immediately, the async operation may start longer than map access time. And the back-end storage memory could be accessed directly via `Map<Key, byte[]>` passed to the back-end constructor (without async call)

src/main/java/com/artipie/asto/memory/InMemoryStorage.java
```diff
@@ -42,7 +44,15 @@
...
-        this.data = new TreeMap<>();
+        this(Collections.emptyMap());
+    }
+
+    /**
+     * Ctor.
+     * @param data Content of storage
+     */
+    public InMemoryStorage(final Map<String, byte[]> data) {
+        this.data = new TreeMap<>(data);
```
g4s8, line 55: This way we can't access passed data by reference: `new TreeMap` calls `putAll()` method on constructor parameter passed, so it just copies all entries from `data` param to internal map. I'd make this constructor package-private and require `NavigableMap<>` isntead of `Map<>` as param, and assign it directly to class field - this way we will be able to access it directly from benchmark storage without sync methods call.

src/test/java/com/artipie/asto/bench/BenchmarkStorageTest.java
```diff
@@ -0,0 +1,93 @@
...
+import org.hamcrest.core.IsEqual;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test for {@link BenchmarkStorage}.
+ * @since 1.1.0
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+final class BenchmarkStorageTest {
```
g4s8, line 23: You forget about the test for `deleted` set

g4s8 (changes requested):
@genryxy looks good for the beginning, please check a few comments below

src/main/java/com/artipie/asto/bench/BenchmarkStorage.java
genryxy: @g4s8 I think that we don't need to keep order for this set, therefore usual set without sorting is more suitable here

src/test/java/com/artipie/asto/bench/BenchmarkStorageTest.java
genryxy: @g4s8 it is impossible now to verify this scenario because `delete` operation is not implemented yet and `deleted` is unreachable from outside

genryxy:
@g4s8 thanks, corrected. Move `BenchmarkStorage` to `memory` package to use a package-private field from `InMemoryStorage`

genryxy:
@g4s8 could you please take a look again

src/main/java/com/artipie/asto/KeyComparator.java
```diff
@@ -0,0 +1,19 @@
...
+package com.artipie.asto;
+
+import java.util.Comparator;
+
+/**
+ * Comparator for key values by their string representation.
+ * @param <T> Type of key
+ * @since 1.1.0
+ */
+public final class KeyComparator<T extends Key> implements Comparator<T> {
```
g4s8, line 14: This class is stateless, so it could be implemented as anonymous class in `Key` static field:
```java
interface Key {
  Comparator<Key> CMP_STRING = (first, second) -> first.string().compareTo(second.string())
}
```
And it doesn't make a sense to use `T` generic which extends `Key`, the comparator interface accepts generic parameters, not provide/return it, so comparators are used as parameter with `? super T` of `Comparator<T>`, e.g. `List<T>:sort` accepts `Comparator<? super T>` to sort the list.

g4s8 (changes requested):
@genryxy please check my comments

genryxy:
@g4s8 thanks, corrected

g4s8 (approved):
@genryxy thanks

PR #319, opened 2021-05-31
Add implementation for `delete` and `exists`

genryxy:
Part of #314
Added implementation of `delete` and  `exists` for benchmark storage and tests. 
For `delete` operation `synchronized` block was used because it is necessary firstly to check existence and after that add to set in case of absence


g4s8 (approved):
@genryxy thanks

PR #320, opened 2021-06-03
Add implementation for `list` and `size`, optimize `delete`

genryxy:
Part of #314
Added implementation of `list` and `size` for benchmark storage and tests.
Optimize `delete` by avoid using synchronize block because the result of `add` operation could be used instead

### DIFF
--- a/src/main/java/com/artipie/asto/memory/BenchmarkStorage.java
+++ b/src/main/java/com/artipie/asto/memory/BenchmarkStorage.java
@@ -69,7 +69,7 @@ public final class BenchmarkStorage implements Storage {
      */
     public BenchmarkStorage(final InMemoryStorage backend) {
         this.backend = backend;
-        this.local = new ConcurrentSkipListMap<>(Key.CMPRTR_STRING);
+        this.local = new ConcurrentSkipListMap<>(Key.CMP_STRING);
         this.deleted = ConcurrentHashMap.newKeySet();
     }
 
@@ -85,7 +85,7 @@ public final class BenchmarkStorage implements Storage {
         return CompletableFuture.supplyAsync(
             () -> {
                 final String prefix = root.string();
-                final Collection<Key> keys = new TreeSet<>(Key.CMPRTR_STRING);
+                final Collection<Key> keys = new TreeSet<>(Key.CMP_STRING);
                 final SortedSet<String> bckndkeys = this.backend.data
                     .navigableKeySet()
                     .tailSet(prefix);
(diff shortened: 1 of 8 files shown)


src/main/java/com/artipie/asto/memory/BenchmarkStorage.java
```diff
@@ -65,8 +81,34 @@ public BenchmarkStorage(final InMemoryStorage backend) {
     }
 
     @Override
-    public CompletableFuture<Collection<Key>> list(final Key prefix) {
-        throw new NotImplementedError("Not implemented yet");
+    public CompletableFuture<Collection<Key>> list(final Key root) {
+        return CompletableFuture.supplyAsync(
+            () -> {
+                final String prefix = root.string();
+                final Collection<Key> keys = new TreeSet<>(Key.CMPRTR_STRING);
```
g4s8, line 88: `CMPRTR_` looks too cryptic ;) Maybe use just `CMP_`?

src/main/java/com/artipie/asto/memory/BenchmarkStorage.java
```diff
@@ -65,8 +81,34 @@ public BenchmarkStorage(final InMemoryStorage backend) {
...
+                        break;
+                    }
+                }
+                for (final Key key : lclkeys) {
+                    if (key.string().startsWith(prefix) && !this.deleted.contains(key)) {
+                        keys.add(key);
+                    } else {
+                        break;
+                    }
+                }
```
g4s8, line 108: Why to implement it manually? You can just use: `addAll` and `removeAll`:
```
set = local.keySet()
set.addAll(backend.keySet()
set.removeAll(deleted)
```

g4s8 (changes requested):
@genryxy please check a few comments

src/main/java/com/artipie/asto/memory/BenchmarkStorage.java
genryxy: @g4s8 there is a test case in `StorageListTest` which would fail in case of using `addAll`
```
...
final byte[] data = "some data!".getBytes();
final BlockingStorage blocking = new BlockingStorage(storage);
blocking.save(new Key.From("1"), data);
blocking.save(new Key.From("a", "b", "c", "1"), data);
blocking.save(new Key.From("a", "b", "2"), data);
blocking.save(new Key.From("a", "z"), data);
blocking.save(new Key.From("z"), data);
final Collection<String> keys = blocking.list(new Key.From("a", "b"))
...
```
It happens because `tailset` returns also keys which are greater than prefix (e.g. `z` and `a/z` in this case)

genryxy:
@g4s8 thanks, renamed and answered

g4s8 (approved):
@genryxy thanks

g4s8 (approved):

PR #321, opened 2021-06-03
Add implementation of `move` and `exclusively`

### LINKED ISSUE #314: Create storage implementation for benchmarks

genryxy:
Closes #314 
Added implementation of `move` and `exclusively`. 
Included `BenchmarkStorage` to `StorageExtension` to enable pool of tests which should be passed by all implementations of storage

### DIFF
--- a/src/main/java/com/artipie/asto/memory/BenchmarkStorage.java
+++ b/src/main/java/com/artipie/asto/memory/BenchmarkStorage.java
@@ -140,18 +140,13 @@ public final class BenchmarkStorage implements Storage {
     public CompletableFuture<Void> move(final Key source, final Key destination) {
         final CompletionStage<Void> res;
         if (this.deleted.contains(source)) {
-            res = artipieIOcompletion("No value for source key", source);
+            res = ioErrorCompletion("No value for source key", source);
         } else {
-            final byte[] lcl = this.local.get(source);
+            final byte[] lcl = this.local.computeIfAbsent(
+                source, key -> this.backend.data.get(key.string())
+            );
             if (lcl == null) {
-                final byte[] bcknd = this.backend.data.get(source.string());
-                if (bcknd == null) {
-                    res = artipieIOcompletion("No value for source key", source);
-                } else {
-                    this.local.put(destination, bcknd);
-                    this.deleted.remove(destination);
-                    res = CompletableFuture.allOf();
-                }
+                res = ioErrorCompletion("No value for source key", source);
             } else {
                 this.local.put(destination, lcl);
                 this.local.remove(source);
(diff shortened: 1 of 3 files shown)


olenagerasimova (dismissed):
@genryxy thanks, looks good to me

genryxy:
@g4s8 could you please check/merge

src/main/java/com/artipie/asto/memory/BenchmarkStorage.java
```diff
@@ -136,7 +138,28 @@ public BenchmarkStorage(final InMemoryStorage backend) {
 
     @Override
     public CompletableFuture<Void> move(final Key source, final Key destination) {
-        throw new NotImplementedError("Not implemented yet");
+        final CompletionStage<Void> res;
+        if (this.deleted.contains(source)) {
+            res = artipieIOcompletion("No value for source key", source);
+        } else {
+            final byte[] lcl = this.local.get(source);
```
g4s8, line 145: You can use `this.local.computeIfAbsent(source, key -> this.backend.data.get(key.string())` instead of these three lines

src/main/java/com/artipie/asto/memory/BenchmarkStorage.java
```diff
@@ -231,4 +249,17 @@ private boolean anyStorageContains(final Key key) {
...
     }
+
+    /**
+     * Obtains failed completion for absent key.
+     * @param msg Text message for exception
+     * @param key Not found key
+     * @param <T> Ignore
+     * @return Failed completion for absent key.
+     */
+    private static <T> CompletionStage<T> artipieIOcompletion(final String msg, final Key key) {
```
g4s8, line 260: I'd rename it by adding error into the name, e.g. `ioErrorCompletion`, etc

g4s8 (changes requested):
@genryxy please check my comments

g4s8 (approved):
@genryxy thanks

PR #324, opened 2021-06-16
Added unchecked scalars and test

### LINKED ISSUE #323: Add unchecked scalars
Let's add unchecked scalars implementations (I mean `Scalar` from cactoos) that will throw `ArtipieException` or `ArtipieIOException`.

olenagerasimova:
Closes #323 
Added unchecked scalars `UncheckedScalar` and `UncheckedIOScalar` and test.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,10 @@ SOFTWARE.
       <version>0.5.4</version>
     </dependency>
     <dependency>
+      <groupId>org.cactoos</groupId>
+      <artifactId>cactoos</artifactId>
+    </dependency>
+    <dependency>
       <groupId>io.etcd</groupId>
       <artifactId>jetcd-test</artifactId>
       <version>0.5.4</version>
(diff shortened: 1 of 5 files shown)


g4s8 (changes requested):
@olenagerasimova I'm not sure if we need to add Cactoos as a compile-time dependency, since it has very unstable API, and previously we had many problems in Artipie with it because of frequently changes in interfaces and method signatures in Cactoos releases, such as `ClassNotFoundException` or `NoSuchMethodException`.

olenagerasimova:
@g4s8 thanks, I see, `Scalar` is very convenient interface though, may we simply create one in `asto`?

g4s8:
> @g4s8 thanks, I see, `Scalar` is very convenient interface though, may we simply create one in `asto`?

@olenagerasimova yes, we can take one stable version to source code

olenagerasimova:
@g4s8 thanks, removed cactoos dependency, copied `Scalar` here

g4s8 (approved):
@olenagerasimova thanks

Issue #325, opened 2021-07-12
Fix dependencies issues


olenagerasimova:
Fuxi reports the following dependencies with high-risk vulnerabilities:
com.fasterxml.jackson.dataformat / jackson-dataformat-cbor / 2.6.7 - test dependency (adobe.testing)
org.apache.ant/ant/1.10.3 - test dependency (org.testing)
org.yaml / snakeyaml / 1.21 - test dependency (org.testing)
com.fasterxml.jackson.core/jackson-databind/2.10.2 - excluded from main code, left in test deps
org.eclipse.jetty/jetty-io/9.4.19.v20190610 - test dependency (io.etcd-test)
org.eclipse.jetty/jetty-webapp/9.4.19.v20190610 - test dependency (io.etcd-test)
org.springframework/spring-web/5.1.10.RELEASE - test dependency (io.etcd-test)
io.vertx/vertx-core/3.9.3 - updated to 4.1.1

Let's check what dependencies are still used and try to get rid of them.

olenagerasimova:
all fixed

PR #326, opened 2021-07-13
Updated vertx, excluded jackson-databind

olenagerasimova:
Part of #325
Updated vertx, checked all deps (see issue descr) and excluded jackson-databind from main code.


g4s8 (approved):
@olenagerasimova thanks

Issue #327, opened 2021-07-13
Copy storage using predicate


g4s8:
Add new constructor to copy items from one storage to another using `Predicate` and not only fixed key's set

PR #328, opened 2021-07-13
feature: copy storage with predicate

g4s8:
Added new construbtor for `Copy` class to copy
storage using `Predicate<? super Key>` instead of fixed set
of keys. Added test.

Issue: #327
Closes: #327


olenagerasimova (approved):
@g4s8 thanks

PR #330, opened 2021-07-19
Exclude fasterxml jackson databind from `vertx-rx-java2`

genryxy:
Part of #325

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.3.2
+
+ - dep: exclude fasterxml jackson databind from vertx-rx-java2
+
 # 1.3.1
 
  - dep: exclude android annotations from etcd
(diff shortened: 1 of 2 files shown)


genryxy:
@g4s8 sorry, I've just noticed that this dependency also excluded from asto in helm and conda, therefore it makes sense to exclude this dependency here

g4s8 (dismissed):
@genryxy thanks. If you need a release, please add this change to the changelog.md file and create a tag with `git tag --annotate vM.m.b`, where M is a major version (same as previous in this case), m - minor (same too), and b - increment +1 from previous. And add description to tag annotation. Then push with `git push --tags origin`

genryxy:
@g4s8 thanks, got it. Updated changelog

g4s8 (approved):
Approved. The tag should be made in merge commit in master branch

Issue #333, opened 2021-09-09
Storage items as io streams


olenagerasimova:
Let's add a mechanism (sdk/classes/methods) to transform storage items to/from input/output streams. Such functionality is necessary as all java libraries to unpack/pack archives, parse/generate xmls/jsons files work with io streams, not with publishers. To implement such mechanism, piped input/output streams and [`rio`](https://github.com/cqfn/rio) will be used. What do we need:
- read item from storage, obtain it as input stream, perform some action (unpack/parse/verify) with this input stream and return some object
- read item from storage, obtain it as input stream, perform some action and write the item back as output stream

It seems that the most correct way to develop convenient and proper sdk here would be to implement functionality step by step and use it in the adapters, making sure it covers all the cases.

PR #334, opened 2021-09-09
Process storage item as input stream

olenagerasimova:
Part of #333 
Added class to process storage item as input stream, this class will be useful in cases when it is necessary to obtain metadata from repository package, usually such operation requires unpacking the archive and reading/parsing some file from it.

### DIFF
--- a/src/main/java/com/artipie/asto/streams/StorageValueAsIStream.java
+++ b/src/main/java/com/artipie/asto/streams/StorageValueAsIStream.java
@@ -1,0 +1,70 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/asto/LICENSE.txt
+ */
+package com.artipie.asto.streams;
+
+import com.artipie.asto.ArtipieIOException;
+import com.artipie.asto.Key;
+import com.artipie.asto.Storage;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+import org.cqfn.rio.WriteGreed;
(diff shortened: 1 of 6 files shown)


olenagerasimova:
@g4s8 please, take a look at the ticked and this first step of implementation, maybe you'll have any ideas to improve.

src/main/java/com/artipie/asto/streams/StorageValueAsIStream.java
```diff
@@ -0,0 +1,70 @@
...
+     * Storage item key.
+     */
+    private final Key key;
+
+    /**
+     * Ctor.
+     * @param asto Abstract storage
+     * @param key Storage item key
+     */
+    public StorageValueAsIStream(final Storage asto, final Key key) {
```
g4s8, line 43: Storage abstraction looks unnecessary high here, we can provide only `Content` to convert it into input stream. This way we can also use this class not only for storage but for HTTP requests too.

src/main/java/com/artipie/asto/streams/StorageValueAsIStream.java
```diff
@@ -0,0 +1,70 @@
...
+import org.cqfn.rio.stream.ReactiveOutputStream;
+
+/**
+ * Process storage value as input stream.
+ * This class allows to treat storage item as input stream and
+ * perform some action with this stream (read/uncompress/parse etc).
+ * @param <T> Result type
+ * @since 1.4
+ */
+public final class StorageValueAsIStream<T> {
```
g4s8, line 26: And I don't really like "value" in the name, it could be `ContentAsStream` (if we wrap content)

g4s8 (changes requested):
@olenagerasimova the general idea looks good, please check my comments

olenagerasimova:
@g4s8 thanks, corrected

g4s8 (approved):
@olenagerasimova thanks

PR #336, opened 2021-09-20
Added class to process asto value as stream

olenagerasimova:
Part of #333 
Added class to read storage value as input stream if exists, process the value and write it back to storage. 
Such workflow is used to create/update repository metadata files (in conda and debian at least).

### DIFF
--- a/src/main/java/com/artipie/asto/streams/StorageValueAsStream.java
+++ b/src/main/java/com/artipie/asto/streams/StorageValueAsStream.java
@@ -1,0 +1,100 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/asto/LICENSE.txt
+ */
+package com.artipie.asto.streams;
+
+import com.artipie.asto.ArtipieIOException;
+import com.artipie.asto.Content;
+import com.artipie.asto.Key;
+import com.artipie.asto.Storage;
+import com.artipie.asto.misc.UncheckedIOConsumer;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.BiConsumer;
+import org.cqfn.rio.Buffers;
+import org.cqfn.rio.WriteGreed;
+import org.cqfn.rio.stream.ReactiveInputStream;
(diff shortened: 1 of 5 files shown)


olenagerasimova:
@g4s8 please, take a look, maybe you'll suggest a better name for the class?

src/main/java/com/artipie/asto/streams/StorageValueAsStream.java
```diff
@@ -0,0 +1,100 @@
...
+import org.cqfn.rio.Buffers;
+import org.cqfn.rio.WriteGreed;
+import org.cqfn.rio.stream.ReactiveInputStream;
+import org.cqfn.rio.stream.ReactiveOutputStream;
+
+/**
+ * Processes storage content as optional input and saves the result back as output stream.
+ * @since 1.5
+ */
+public final class StorageValueAsStream {
```
g4s8, line 30: The name looks a bit confusing because I'm expecting this class to be an implementation of `Stream`. Maybe we can choose a name to say that this class connects input and output, maybe something like `pipeline` or similar

src/main/java/com/artipie/asto/streams/StorageValueAsStream.java
```diff
@@ -0,0 +1,100 @@
...
+                    } else {
+                        tmp = CompletableFuture.allOf();
+                        oinput = Optional.empty();
+                    }
+                    final PipedInputStream src = new PipedInputStream(resout);
+                    future = tmp.thenCompose(
+                        nothing -> this.asto.save(
+                            this.key,
+                            new Content.From(
+                                new ReactiveInputStream(src).read(Buffers.Standard.K16)
```
g4s8, line 85: Why 16Kb here? I'd suggest using default system 8Kb (as a standard system buffer size for disk read operation)

src/test/java/com/artipie/asto/streams/ProcessContentTest.java
```diff
@@ -0,0 +1,87 @@
...
+                try {
+                    IOUtils.write(text, out, charset);
+                } catch (final IOException err) {
+                    throw new ArtipieIOException(err);
+                }
+            }
+        ).toCompletableFuture().join();
+        MatcherAssert.assertThat(
+            "test.txt does not contain text `Hello world!`",
+            new PublisherAs(this.asto.value(key).join()).asciiString().toCompletableFuture().join(),
```
g4s8, line 82: You can use `BlockingStorage` for tests: `new String(new BlockingStorage(this.ast).value(key))`

g4s8 (changes requested):
@olenagerasimova please check my comments

olenagerasimova:
@g4s8 thanks, all fixed

PR #338, opened 2021-09-27
Added `StorageValuePipelineResult` with function

olenagerasimova:
Part of #333 
Added class to process storage value as IO streams and return some value.


olenagerasimova:
@g4s8 could you please take a look? Is it possible to avoid code duptications?

g4s8 (commented):
@olenagerasimova what is the use case of this class? It looks very similar to previous one for stream processing

olenagerasimova:
@g4s8 yes, the only difference is that it returns value, not void. The example when it's necessary is in debian: when repository metadata file is updated, we need to check for duplicate packages and then remove them. Duplicates is returned while processing the repodata, so consumer is not applicable in this case, we need function.

g4s8:
@olenagerasimova then, maybe we can use same logic in both places? E.g. the class may have two methods: one for returning `<T>` and another for `void`, and second method can use first and ignoring element

olenagerasimova:
@g4s8 I was trying to do so, but it led to the following problems:
a) as I will have to parametrize the class, this parameter is useless for the method returning void
b) method, returning void should accept `BiConsumer` and return `void`, in this case there is no way (or at least I do not see any) to determine type for the parameter
c) also, having methods with `BiConsumer` and `BiFunction` potentially leads to ambiguous call problems

olenagerasimova:
@g4s8 please, take a look, I already need this new method with function in two adapters...

g4s8 (approved):
@olenagerasimova thanks

Issue #218, opened 2020-08-05
Class to obtain test resource 


olenagerasimova:
Let's introduce class to obtain resource from current class loader and put this resource to provided storage. 

This functionality is required in every adapter, this class will allow to avoid code duplications and unexpected errors.

olenagerasimova:
@0crat assign me

olenagerasimova:
@rultor release, tag=`0.23.4`

PR #339, opened 2021-09-30
feat: test resource save to file

g4s8:
Added method for `TestResource` to save it directly to
file (`Path`).

Ticket: #218
Related: artipie/gem-adapter#130


olenagerasimova (approved):
@g4s8 thanks

PR #340, opened 2021-10-13
Test with `StorageValuePipeline` and `CompletableFuture.allOff()`

olenagerasimova:
Part of #333 
Added one test for `StorageValuePipeline`


olenagerasimova:
not valid

PR #342, opened 2021-10-15
feat: adds UncheckedSupplier

baudoliver7:
For artipie/gem-adapter#101 :

- Adds `UncheckedSupplier` and `UncheckedSupplier` classes
- Introduces unit tests for both classes


olenagerasimova:
@g4s8 please, check/merge

PR #347, opened 2021-10-29
feat: Excludes any part from storage key

### LINKED ISSUE #341: Exclude any part from storage key
Let's add class to exclude any part (by part string value or index) from the existing key and create a new key without this excluded part. By a part I mean a string between `/` chars, a part cannot contain separator char `\`. 
Let's suppose the required class is called `KeyParts`, here is an example:
```java
// exclude part by part value
new KeyParts(new Key.From("one", "two")).exclude("one") // result is Key instance new Key.From("two")
new KeyParts(new Key.From("one", "two")).exclude("two") // result is Key instance new Key.From("one")
// exclude part by index (starting with 0)
new KeyParts(new Key.From("one", "two")).exclude(0) // result is Key instance new Key.From("two")
new KeyParts(new Key.From("one", "two")).exclude(1) // result is Key instance new Key.From("one")

// `exclude` should exclude all found parts
new KeyParts(new Key.From("one", "two", "one", "three")).exclude("one") // result is Key instance new Key.From("two", "three")
// we can also add `excludeFirst` and `excludeLast` to exclude first/last part
new KeyParts(new Key.From("one", "two", "one", "three")).excludeFirst("one") // result is Key instance new Key.From("two", "one", "three")
```
If part to exclude was not found (or index is out of bounds), the class can return the origin key. Also, this same class can be extended with methods to append the part in the middle of the key, correspondingly to `List.add(object, index)` method, get key parts as list or stream etc.

baudoliver7:
Close #341 

- Introduces `parts` method in `Key`
- Introduces classes `KeyExcludeFirst` and `KeyExcludeAll`
- Introduces tests for them
- Adds a new puzzle for other operations

### DIFF
--- a/src/main/java/com/artipie/asto/Key.java
+++ b/src/main/java/com/artipie/asto/Key.java
@@ -194,7 +194,7 @@ public interface Key {
 
         @Override
         public List<String> parts() {
-            return this.parts.stream().collect(Collectors.toList());
+            return Collections.unmodifiableList(this.parts);
         }
 
         @Override


olenagerasimova (approved):
@baudoliver7 thanks, looks good

src/main/java/com/artipie/asto/Key.java
```diff
@@ -18,6 +19,10 @@
  * Storage key.
  *
  * @since 0.6
+ * @todo #341:30min Add others key operations (exclusion by index, insertion,..).
+ *  We should exclude a part from a storage key by index. We should also insert
+ *  a part to a key by specifying the position (Like we do with {@code List.add(Obj, index)}).
+ *  We also want to get parts of a storage key as list or stream.
```
g4s8, line 25: @olenagerasimova I'm not sure if it's the part of origin ticket, isn't it? @baudoliver7 if not, please create a new ticket and remove this `todo`

baudoliver7: @olenagerasimova just to say we are still waiting for your answer before completing this ticket :)

olenagerasimova: @baudoliver7 @g4s8 I think this is part of the ticket, I've provided corresponding examples in the description

baudoliver7: @olenagerasimova Thanks :)

olenagerasimova:
@g4s8 please, take a look/merge

g4s8 (changes requested):
@baudoliver7 looks good, but I think `Key` class may become too big with a lot of nested classes. I'd suggest moving new classes to a new package: `com.artipie.key.KeyExcludeAll` and `com.artipie.key.KeyExcludeFirst`

baudoliver7:
@g4s8 please, `com.artipie.key` or `com.artipie.asto.key` ?

Do you want me to move also `Key` class in that package ?

g4s8:
> @g4s8 please, `com.artipie.key` or `com.artipie.asto.key` ?
> 
> Do you want me to move also `Key` class in that package ?

@baudoliver7 no, `Key` is stable interface, no need to change it, move only new classes.

baudoliver7:
@g4s8 I moved the new classes in `com.artipie.asto.key` because `Key` class is in `com.artipie.asto`. Is it ok ?

src/main/java/com/artipie/asto/key/KeyParts.java
```diff
@@ -0,0 +1,46 @@
...
+import com.artipie.asto.Key;
+import org.cactoos.list.ListEnvelope;
+import org.cactoos.list.ListOf;
+
+/**
+ * Parts of a key.
+ *
+ * @since 1.8.1
+ */
+public final class KeyParts extends ListEnvelope<String> {
```
g4s8, line 16: I'm not sure that we need a separate class for this. Maybe we can add a method to `Key` (because each key implementation could be split in parts, and the delimiter is a part of contract). @artipie/members WDYT?

g4s8 (changes requested):
@baudoliver7 please check my comments

src/main/java/com/artipie/asto/key/KeyParts.java
baudoliver7: @g4s8 I was thinking about that ...

pom.xml
```diff
@@ -171,6 +171,12 @@ SOFTWARE.
         </exclusion>
       </exclusions>
     </dependency>
+    <!-- https://mvnrepository.com/artifact/org.cactoos/cactoos -->
+    <dependency>
+      <groupId>org.cactoos</groupId>
+      <artifactId>cactoos</artifactId>
+      <version>0.50</version>
```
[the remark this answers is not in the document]
baudoliver7: @g4s8 ok

src/main/java/com/artipie/asto/key/KeyParts.java
baudoliver7: @artipie/members I think it will be nice and useful to add to `Key` interface this :
```java
public interface Key {

    /**
     * Delimiter used to split string into parts and join parts into string.
     */
    String DELIMITER = "/";
    
    /**
     * Parts of key.
     * @return List of parts
     */
    List<String> parts();
```

@olenagerasimova WDYT ?

baudoliver7:
@g4s8 please, needs check/merge :)

baudoliver7:
@g4s8 ping

src/main/java/com/artipie/asto/Key.java
```diff
@@ -178,6 +192,11 @@ public String string() {
             return parent;
         }
 
+        @Override
+        public List<String> parts() {
+            return this.parts.stream().collect(Collectors.toList());
```
g4s8, line 197: You can just use `return Collections.unmodifiableList(this.parts)` without creating a stream here.

g4s8 (changes requested):
@baudoliver7 almost good, just final minor comment

baudoliver7:
@g4s8 If everything is ok, could you merge ?

g4s8 (approved):
@baudoliver7 thanks

PR #349, opened 2021-11-02
Extended `StorageValuePipeline` to accept diff keys

olenagerasimova:
Part of #333 
Extended `StorageValuePipeline` to accept diff keys: one key is to read from and another is to write processing result to.

This is needed in rpm to read metadata from repository location and save the result to temp location for further processing.


genryxy (approved):
@olenagerasimova thanks, looks good to me

olenagerasimova:
@g4s8 please, merge

Issue #346, opened 2021-10-29
New metadata method for storage


g4s8:
# Problem

Some storage implementation provides additional metadata for blobs, which can be used for fast items processing: e.g. S3 storage can provide difference hash-sums of blob with `HEAD` request, so it could be used for fast item upload verification in many adapters instead of manual hash-sum calculation; also, almost all adapters can provide create/update time and blob size.

# Proposal

Add `metadata(Key)` method to `Storage` interface to read metadata of blob item. It could include size too, so `Storage.size` could be deprecated.

g4s8:
> Oops, it's reserved for @artipie/members ! All my apologies.

@baudoliver7 actually I wanted to hear opinions of all participants who works with this code, so your comment is helpful too.

PR #350, opened 2021-11-03
feat(metadata): storage metadata API

g4s8:
Added `Meta` interface to represent blob metadata,
updated storage with new method `metadata(Key)`,
deprecated `size(Key)` method since it's covered by metadata method.
Added interfaces `Meta.ReadOperator<T>` and `Meta.WriteOperator<T>`
to serialize and parse data values to intermediate `Map<String, String>`
container. Updated all `Storage`s implementation and related tests.

**Reasons behind this implementation:** each storage can provide different
metadata, e.g. storage based on file-system `FileStorage` can provide
timestampts for file creation time, update time and access time;
S3 storage can provide MD5 hash of content without loading it from
remote bucket, etc. So we need some dynamic and extendable metadata
representation, `Map<String, String>` could be good for this. Another
problem to solve - storage API may be different - so most probably we
always need to check existance of some metadata fields here.
Also, we may want to select custom tuples of metadata from intermediate
representation, so we need some flexible API to read it.

**Solution:** `Meta` interface with single abstract method `<T> T
read(ReadOperator<T>)` was introduced to query metadata object.
`Meta` implementation depends on storage internals, e.g. for S3
it encapsulates `HeadObjectResponse` for FS - `BasicFileAttributes`.
Also, added `WriteOperator<T>` and simple implementation of both
interfaces to encapsulate metadata key and use same object for reading
and writing. `ReadOperator` could be implemented by client to combine
multiple operators into tuples to merge metadata results.
Write operators could be added later with bckward compatibility without
affecting current code.

Closes: #346

### DIFF
--- a/src/main/java/com/artipie/asto/Meta.java
+++ b/src/main/java/com/artipie/asto/Meta.java
@@ -1,0 +1,155 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/asto/LICENSE.txt
+ */
+package com.artipie.asto;
+
+import java.time.Instant;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ * Storage content metadata.
+ * @since 1.9
+ */
+public interface Meta {
+
+    /**
+     * Operator for MD5 hash.
+     */
+    OpRWSimple<String> OP_MD5 = new OpRWSimple<>("md5", Function.identity());
+
+    /**
+     * Operator for size.
+     */
+    OpRWSimple<Long> OP_SIZE = new OpRWSimple<>("size", Long::parseLong);
+
+    /**
+     * Operator for created time.
+     */
+    OpRWSimple<Instant> OP_CREATED_AT = new OpRWSimple<>("created-at", Instant::parse);
+
+    /**
+     * Operator for updated time.
+     */
+    OpRWSimple<Instant> OP_UPDATED_AT = new OpRWSimple<>("updated-at", Instant::parse);
+
+    /**
+     * Operator for last access time.
+     */
+    OpRWSimple<Instant> OP_ACCESSED_AT = new OpRWSimple<>("accessed-at", Instant::parse);
+
+    /**
+     * Empty metadata.
+     */
+    Meta EMPTY = new Meta() {
+        @Override
+        public <T> T read(final Meta.ReadOperator<T> opr) {
+            return opr.take(Collections.emptyMap());
+        }
+    };
+
+    /**
+     * Read metadata.
+     * @param opr Operator to read
+     * @param <T> Value type
+     * @return Metadata value
+     */
+    <T> T read(Meta.ReadOperator<T> opr);
+
+    /**
+     * Metadata read operator.
+     * @param <T> Result type
+     * @since 1.0
+     */
+    @FunctionalInterface
+    interface ReadOperator<T> {
+
+        /**
+         * Take metadata param from raw metadata.
+         * @param raw Readonly map of strings
+         * @return Metadata value
+         */
+        T take(Map<String, ? extends String> raw);
+    }
+
+    /**
+     * Metadata write operator.
+     * @param <T> Value type
+     * @since 1.9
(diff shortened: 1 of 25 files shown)


g4s8:
@artipie/members please check this PR

src/main/java/com/artipie/asto/Meta.java
```diff
@@ -0,0 +1,155 @@
...
+        void put(Map<String, String> raw, T val);
+    }
+
+    /**
+     * Read and write simple operator implementation.
+     * @param <T> Value type
+     * @since 1.9
+     * @checkstyle AbbreviationAsWordInNameCheck (5 lines)
+     */
+    final class OpRWSimple<T> implements ReadOperator<Optional<? extends T>>, WriteOperator<T> {
```
olenagerasimova, line 100: I'd suggest to name the class `SimpleRWOp` - simple read/write operator. "Operator read/write simple" sound upside-down

src/main/java/com/artipie/asto/Storage.java
```diff
@@ -62,8 +62,19 @@
      *
      * @param key The key of value.
      * @return Size of value in bytes.
+     * @deprecated Use {@link #metadata(Key)} to get content size
      */
-    CompletableFuture<Long> size(Key key);
+    @Deprecated
```
olenagerasimova, line 67: hm, with this default implementation, what is the real reason to deprecate this method? We will repeat the same code in the adapters...

src/test/java/com/artipie/asto/fs/FileMetaTest.java
```diff
@@ -0,0 +1,53 @@
...
+ * @since 1.9
+ */
+final class FileMetaTest {
+    @Test
+    void readAttrs() {
+        final long len = 4;
+        final Instant creation = Instant.ofEpochMilli(1);
+        final Instant modified = Instant.ofEpochMilli(2);
+        final Instant access = Instant.ofEpochMilli(3);
+        final BasicFileAttributes attrs = Mockito.mock(BasicFileAttributes.class);
```
olenagerasimova, line 27: is not it possible to create `BasicFileAttributes` instance without mockito?

olenagerasimova (changes requested):
@g4s8 thanks, check my comments please

src/main/java/com/artipie/asto/Storage.java
g4s8: The reason for making it `default` is to keep backward compatibility, we'll be able to update this dependency in `artipie/artipie` and keeping all adapters working. The reason of deprecation is to require all adapters eventually replace usage of this method with new one due to compilation warning. So later we'll remove it safely.

src/test/java/com/artipie/asto/fs/FileMetaTest.java
g4s8: It's quite big interface with many attributes without public implementations (internal implementations depends on OS, e.g. `UnixFileAttributes` and `DosFileAttributes`). So instead of making fake classes with more than 10 methods (where we need only 3 methods) it looks easier to make mock in this case.

src/main/java/com/artipie/asto/Meta.java
g4s8: Agree, will fix it.

src/main/java/com/artipie/asto/Storage.java
olenagerasimova: Sorry, seems like I do not express my point clear enough. I do not think, that this default implementation should be deprecated, because to obtain the size of the storage item in the adapters we will repeat the same code, that is written in this default implementation.

g4s8: I think we have to avoid duplicating functionality in one place, without deprecation it would be possible to use two different methods of `Storage` to get the size. If using `meta` method would be not comfortable, we can introduce additional helper object with frequently used operations, e.g. `long size = new MetaAttributes(meta).size()`, WDYT?

olenagerasimova: ok, sounds good

olenagerasimova (approved):
@g4s8 thanks

PR #353, opened 2021-11-08
Fixed and tested `SubStorage`

### LINKED ISSUE #351: List items of `SubStorage`
I found two problems with `SubStorage`:
1) Complete lack of tests
2) `list` operation does work correctly in some cases: when `SubStorage` is created with `Ket.ROOT` as prefix, `list` removes first met `/` symbol, thus returning not existing items. For example, if the sub storage contains `abc/xyz/hello.txt`, then list by `abc` key will return `abcxyz/hello.txt`

olenagerasimova:
Closes #351 
Fixed problem with `list` operation of `SubStorage`, added some tests, todo for other tests.

The problem is described in the original issue:
> I found two problems with SubStorage:
> 1) Complete lack of tests
> 2) `list` operation does work correctly in some cases: when `SubStorage` is created with `Ket.ROOT` as prefix, list removes first met `/` symbol, thus returning not existing items. For example, if the sub storage contains `abc/xyz/hello.txt`, then list by `abc` key will return `abcxyz/hello.txt`
> 

The first problem is fixed by adding some tests (complete set of tests will be added later, by the `todo`), the second problem is fixed by correcting the regexp, used in `list` method.

### DIFF
--- a/src/main/java/com/artipie/asto/SubStorage.java
+++ b/src/main/java/com/artipie/asto/SubStorage.java
@@ -53,7 +53,7 @@ public final class SubStorage implements Storage {
 
     @Override
     public CompletableFuture<Collection<Key>> list(final Key filter) {
-        final Pattern ptn = Pattern.compile(String.format("%s/", this.prefix.string()));
+        final Pattern ptn = Pattern.compile(String.format("^%s/", this.prefix.string()));
         return this.origin.list(new PrefixedKed(this.prefix, filter)).thenApply(
             keys -> keys.stream()
                 .map(key -> new Key.From(ptn.matcher(key.string()).replaceFirst("")))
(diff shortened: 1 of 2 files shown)


g4s8:
@olenagerasimova could you please describe what was the problem, and how did you fix it in PR?

olenagerasimova:
@g4s8 done, the problems/bugs ets are supposed to described in the issues, do not they?

olenagerasimova:
@g4s8 please, review, I need this fix to continue the work

g4s8:
> @g4s8 done, the problems/bugs ets are supposed to described in the issues, do not they?

@olenagerasimova you are right, only solution is needed for PR

g4s8 (approved):
@olenagerasimova thanks

PR #366, opened 2021-11-15
test: Adds some tests for SubStorage

### LINKED ISSUE #354: Add more tests for {@link SubStorage}: all the methods of the class should be verified, do not...
The puzzle `351-e3dca535` from #351 has to be resolved: 

https://github.com/artipie/asto/blob/90387fff8316c8faa288c5288a7b753e881ad8fb/src/test/java/com/artipie/asto/SubStorageTest.java#L20-L22

The puzzle was created by @olenagerasimova on 09-Nov-21. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

baudoliver7:
Closes #354

- Test methods `size`, `move` and `save`
- Add a puzzle to continue to add more tests

### DIFF
--- a/src/test/java/com/artipie/asto/SubStorageTest.java
+++ b/src/test/java/com/artipie/asto/SubStorageTest.java
@@ -109,14 +109,14 @@ class SubStorageTest {
 
     @ParameterizedTest
     @ValueSource(strings = {"my-project", "com/example"})
-    void savesValue(final String pref) {
+    void savesContent(final String pref) {
         final Key prefix = new Key.From(pref);
         final byte[] data = "some data".getBytes(StandardCharsets.UTF_8);
         final SubStorage substo = new SubStorage(prefix, this.asto);
         substo.save(new Key.From("package"), new Content.From(data)).join();
         MatcherAssert.assertThat(
             "Returns storage item with prefix",
-            new BlockingStorage(substo).value(new Key.From("package")),
+            new BlockingStorage(this.asto).value(new Key.From(prefix, "package")),
             new IsEqual<>(data)
         );
         MatcherAssert.assertThat(
@@ -129,7 +129,7 @@ class SubStorageTest {
 
     @ParameterizedTest
     @ValueSource(strings = {"my-project", "com/example"})
-    void movesValue(final String pref) {
+    void movesContent(final String pref) {
         final Key prefix = new Key.From(pref);
         final byte[] data = "source".getBytes(StandardCharsets.UTF_8);
         final Key source = new Key.From("src");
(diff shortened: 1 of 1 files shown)


baudoliver7:
@artipie/maintainers needs a review :)

src/test/java/com/artipie/asto/SubStorageTest.java
```diff
@@ -106,4 +106,73 @@ void checksExistence(final String pref) {
             new SubStorage(Key.ROOT, this.asto).exists(new Key.From(prefix, "any.txt")).join()
         );
     }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"my-project", "com/example"})
+    void savesValue(final String pref) {
```
olenagerasimova, line 112: I'm not sure this test method name and correspond to each other. To test `save` method of the `SubStorage` seems logical to 
1) create instance of the `SubStorage` with different prefixes
2) call `save` method of these `SubStorage` instances
3) check that content is saved and the key is correct by getting the value from the original storage by prefixed key. Thus, assuming, that original storage works correctly, we can be sure that: a) content is stored with the expected key b) content is correct

src/test/java/com/artipie/asto/SubStorageTest.java
```diff
@@ -106,4 +106,73 @@ void checksExistence(final String pref) {
...
+            new Key.From(prefix, destination),
+            new Content.From("destination".getBytes(StandardCharsets.UTF_8))
+        ).join();
+        final SubStorage substo = new SubStorage(prefix, this.asto);
+        substo.move(source, destination).join();
+        MatcherAssert.assertThat(
+            "Moves key value with prefix",
+            new BlockingStorage(substo).value(destination),
+            new IsEqual<>(data)
+        );
```
olenagerasimova, line 148: About the same idea here: use original storage in the assertion, `SubStorage` is being tested here, we do not know if any method works correctly

olenagerasimova (changes requested):
@baudoliver7 thanks, please, check my comments

src/test/java/com/artipie/asto/SubStorageTest.java
baudoliver7: olenagerasimova Got it :)

olenagerasimova (approved):
@baudoliver7 thanks
@g4s8 please, take a look/merge

PR #368, opened 2021-11-16
feat: introduce a default method in Storage to delete prefixed keys

### LINKED ISSUE #348: Remove items from the storage by prefix key
Very often it's necessary to remove items from storage by some key prefix:
```
private CompletableFuture<Void> remove(final Key key) {
        return this.asto.list(key).thenCompose(
            list -> CompletableFuture.allOf(
                list.stream().map(this.asto::delete)
                    .toArray(CompletableFuture[]::new)
            )
        );
    }
```
Such methods are duplicated a lot in different adapters. I'd suggest to add class for this operation
(the example can be improved by checking that item exists before performing remove operation).

baudoliver7:
Closes #348

- Introduce a default method `deleteAll` to `Storage`
- Introduce a test for this method


baudoliver7:
@artipie/maintainers need a review here :)

g4s8 (changes requested):
@baudoliver7 @olenagerasimova I see two potentical problems here:
 * some storages may have fast remove operation by prefix support, need to check if storages has APIs for batch-remove operations. In that case this method should be the part of `Storage` interface and implemented on lower level.
 * all keys are removed simultaneously - it may cause traffic overload on huge amount of keys to remove, server may reset connection in this case and we'll get partial batch removing and "reset connection by peer" error for the rest (need to check if scheduler sends requests by chunks)

olenagerasimova:
@g4s8 hm, then, implementing this operation on `Storage` level seems to be the best solution as there we know how exactly operation is performed and can deal accordingly.

g4s8:
@baudoliver7 I'd suggest adding current implementation as default method to `Storage` interface and todo for implementing this method in storage implementations. @olenagerasimova WDYT?

olenagerasimova:
@g4s8 yes, sound reasonable

baudoliver7:
@g4s8 @olenagerasimova Ok, let's me apply it :)

baudoliver7:
@g4s8 @olenagerasimova please, could you take a look to these new changes ?

src/test/java/com/artipie/asto/StorageSaveAndLoadTest.java
```diff
@@ -104,8 +104,8 @@ void shouldSaveEmpty(final Storage storage) throws Exception {
         storage.save(key, new Content.OneTime(new Content.From(Flowable.empty()))).get();
         MatcherAssert.assertThat(
             "Saved content should be empty",
-            new BlockingStorage(storage).value(key),
-            Matchers.equalTo(new byte[0])
+            new BlockingStorage(storage).value(key).length,
+            Matchers.equalTo(0)
```
g4s8, line 108: This change is not valid

src/main/java/com/artipie/asto/blocking/BlockingStorage.java
```diff
@@ -108,4 +108,13 @@ public long size(final Key key) {
     public void delete(final Key key) {
         this.storage.delete(key).join();
     }
+
+    /**
+     * Removes all items with key prefix.
+     *
+     * @param prefix Key prefix.
+     */
+    public void deleteAll(final Key prefix) {
```
g4s8, line 117: Also, please update `LoggingStorage` in the same way

src/main/java/com/artipie/asto/Storage.java
```diff
@@ -99,6 +99,21 @@
...
+     * @param prefix Key prefix.
+     * @return Completion or error signal.
+     */
+    default CompletableFuture<Void> deleteAll(final Key prefix) {
+        return this.list(prefix).thenCompose(
+            list -> CompletableFuture.allOf(
+                list.stream().map(this::delete)
+                    .toArray(CompletableFuture[]::new)
+            )
+        );
```
olenagerasimova, line 114: I'm not sure if it's really necessary, but I'd suggest adding a check that key exists before remove it

src/test/java/com/artipie/asto/StorageDeleteAllTest.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @since 1.9.1
+ * @todo #348:30min Fix error when trying to save data with prefixed Key.
+ *  Test below goes in error when we try to save data in {@link com.artipie.asto.fs.FileStorage}
+ *  and {@link com.artipie.asto.fs.VertxFileStorage} with prefixed key, after saving data with
+ *  key prefix. Please, fix it and enable this test.
+ */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+@ExtendWith(StorageExtension.class)
+public final class StorageDeleteAllTest {
```
olenagerasimova, line 25: And here I'd suggest to add test to remove by `Key.ROOT`

olenagerasimova (changes requested):
@baudoliver7 thanks, please, check my comment

src/test/java/com/artipie/asto/StorageDeleteAllTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * Tests for {@link Storage#deleteAll(Key)}.
+ *
+ * @since 1.9.1
+ * @todo #348:30min Fix error when trying to save data with prefixed Key.
+ *  Test below goes in error when we try to save data in {@link com.artipie.asto.fs.FileStorage}
+ *  and {@link com.artipie.asto.fs.VertxFileStorage} with prefixed key, after saving data with
+ *  key prefix. Please, fix it and enable this test.
```
g4s8, line 21: This todo is not valid - it's not possible to store data in storage with prefixes in file-system: you have file `one` and then you're trying to save file `one/two` using `one` as directory, so it's not possible to fix such scenario.

g4s8 (changes requested):
@baudoliver7 please check my comments

src/test/java/com/artipie/asto/StorageDeleteAllTest.java
baudoliver7: @g4s8 Ok

src/test/java/com/artipie/asto/StorageDeleteAllTest.java
baudoliver7: @olenagerasimova Ok

src/main/java/com/artipie/asto/Storage.java
baudoliver7: @olenagerasimova I think It's not necessary since if the key doesn't exist, `list` will not return it.

g4s8 (approved):
@baudoliver7 thanks!

g4s8:
@olenagerasimova please check your review fixes & answers

baudoliver7:
@olenagerasimova waiting your review :)

olenagerasimova (approved):
@g4s8 @baudoliver7 thanks, looks good to merge

baudoliver7:
@g4s8 looks like there is a trouble on windows :-( it's really random !
I don't know why, perhaps about text encoding ?

baudoliver7:
@g4s8 There was some trouble here during merging.

baudoliver7:
@g4s8 please, could try to merge now ?

g4s8 (approved):

baudoliver7:
@g4s8 what do you use as default encoding locally ? `us-ascii` or `utf-8` ?

g4s8:
> @g4s8 what do you use as default encoding locally ? `us-ascii` or `utf-8` ?

@baudoliver7 what is the use case?

baudoliver7:
@g4s8 We got a trouble on `StorageDeleteAllTest.java` when merging. I think it is due to its file encoding. I use locally `us-ascii`. Am I doing well ?

PR #373, opened 2021-11-18
Compare keys by string representation

### LINKED ISSUE #371: SubStorage: list by `Key.ROOT`
`SubStorage`, created with `Key.ROOT`, fails to list items by `Key.ROOT` with `ArtipieException: Empty parts are not allowed`, it can be reproduced with `FileStorage` storage:
```
new SubStorage(Key.ROOT, new FileStorage(temp)).list(Key.ROOT).join();
```

In this case, `SubStorage` should properly handle empty parts of the key and the result should be the same as if we list items by `Key.ROOT` of the origin storage.

olenagerasimova:
Closes #371 
The problem is described in #371, but what is actually wrong is to call equals on keys like
`Key.ROOT.equals(anotherKey)` 
as in the case with `SubStorage` `anotherKey` is an instance of `SubStorage.PrefixedKey`, not `Key.From` and `equals()` (this is the `equals()` overridden in Key.From` class) will return false.

I've fixed it by comparing keys by their string representations, added various `SubStorage`'s to `StorageExtension` and test with `Key.ROOT` to `StorageListTest`.

### DIFF
--- a/src/main/java/com/artipie/asto/fs/FileStorage.java
+++ b/src/main/java/com/artipie/asto/fs/FileStorage.java
@@ -201,7 +201,7 @@ public final class FileStorage implements Storage {
     @Override
     public CompletableFuture<Content> value(final Key key) {
         final CompletableFuture<Content> res;
-        if (Key.ROOT.equals(key)) {
+        if (Key.ROOT.string().equals(key.string())) {
             res = new CompletableFutureSupport.Failed<Content>(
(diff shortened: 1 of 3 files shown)


genryxy (approved):
@olenagerasimova thanks, looks good to me
@g4s8 please check

olenagerasimova:
@g4s8 please, take a look

Issue #309, opened 2021-04-21
The test is disabled, the CI can't start ETCD extension which depends on testcontainer with etcd...


0pdd:
The puzzle `306-ccf10929` from #306 has to be resolved: 

https://github.com/artipie/asto/blob/fc2ec934d3028b695a17e487f4a368fa425bac06/src/test/java/com/artipie/asto/etcd/EtcdStorageITCase.java#L46-L48

The puzzle was created by @g4s8 on 21-Apr-21. 

Estimate: 90 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@0pdd the puzzle [#384](https://github.com/artipie/asto/issues/384) is still not solved.

0pdd:
@0pdd the puzzle [#unknown]() is still not solved.

PR #377, opened 2021-11-22
test: setup etcd cluster for StorageExtension

### LINKED ISSUE #310: Add etcd storage to StorageExtension to run all common tests on EtcdStorage too. It could be not...
The puzzle `306-f94460d8` from #306 has to be resolved: 

https://github.com/artipie/asto/blob/fc2ec934d3028b695a17e487f4a368fa425bac06/src/test/java/com/artipie/asto/etcd/EtcdStorageITCase.java#L49-L52

The puzzle was created by @g4s8 on 21-Apr-21. 

Estimate: 90 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

baudoliver7:
Closes #310

- Setup etcd cluster for `StorageExtension`
- Fix some tests
- Disable `EtcdStorage` tests for `Windows` in `StorageExtension` tests

### DIFF
--- a/src/main/java/com/artipie/asto/SubStorage.java
+++ b/src/main/java/com/artipie/asto/SubStorage.java
@@ -53,8 +53,14 @@ public final class SubStorage implements Storage {
 
     @Override
     public CompletableFuture<Collection<Key>> list(final Key filter) {
+        final Key cleanfilter;
+        if (this.prefix == Key.ROOT && filter == Key.ROOT) {
+            cleanfilter = Key.ROOT;
+        } else {
+            cleanfilter = new PrefixedKed(this.prefix, filter);
+        }
         final Pattern ptn = Pattern.compile(String.format("^%s/", this.prefix.string()));
-        return this.origin.list(new PrefixedKed(this.prefix, filter)).thenApply(
+        return this.origin.list(cleanfilter).thenApply(
             keys -> keys.stream()
                 .map(key -> new Key.From(ptn.matcher(key.string()).replaceFirst("")))
                 .collect(Collectors.toList())
--- a/src/main/java/com/artipie/asto/SubStorage.java
+++ b/src/main/java/com/artipie/asto/SubStorage.java
@@ -53,14 +53,8 @@ public final class SubStorage implements Storage {
 
     @Override
     public CompletableFuture<Collection<Key>> list(final Key filter) {
-        final Key cleanfilter;
-        if (this.prefix == Key.ROOT && filter == Key.ROOT) {
-            cleanfilter = Key.ROOT;
-        } else {
-            cleanfilter = new PrefixedKed(this.prefix, filter);
-        }
         final Pattern ptn = Pattern.compile(String.format("^%s/", this.prefix.string()));
-        return this.origin.list(cleanfilter).thenApply(
+        return this.origin.list(new PrefixedKed(this.prefix, filter)).thenApply(
             keys -> keys.stream()
                 .map(key -> new Key.From(ptn.matcher(key.string()).replaceFirst("")))
                 .collect(Collectors.toList())
--- a/src/test/java/com/artipie/asto/StorageExtension.java
+++ b/src/test/java/com/artipie/asto/StorageExtension.java
@@ -125,12 +125,15 @@ final class StorageExtension
                 )
             );
             if (!SystemUtils.IS_OS_WINDOWS) {
+                storages.add(this.etcdStorage());
                 storages.add(
-                    new EtcdStorage(
-                        Client.builder()
-                            .endpoints(this.etcd.getClientEndpoints())
-                            .build()
+                    new SubStorage(
+                        new Key.From("etcd-prefix"),
+                        this.etcdStorage()
                     )
+                );
+                storages.add(
+                    new SubStorage(Key.ROOT, this.etcdStorage())
                 );
             }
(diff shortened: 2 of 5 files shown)


baudoliver7:
@g4s8 To continue on this issue, I need to resolve issue #309 first.
Please, could you assign it to me ?

olenagerasimova (approved):
@baudoliver7 thanks, looks good to me

baudoliver7:
@olenagerasimova thanks :)
@g4s8 please, could you check/merge ?

baudoliver7:
@g4s8 ping

baudoliver7:
@g4s8 please, could you check/merge ?

src/test/java/com/artipie/asto/StorageExtension.java
```diff
@@ -78,21 +108,33 @@ public boolean supportsTestTemplate(final ExtensionContext context) {
...
+                    new SubStorage(Key.ROOT, new InMemoryStorage()),
+                    new SubStorage(Key.ROOT, new FileStorage(Files.createTempDirectory("sub"))),
+                    new FileStorage(Files.createTempDirectory("junit")),
+                    new VertxFileStorage(Files.createTempDirectory("vtxjunit"), this.vertx),
+                    new BenchmarkStorage(new InMemoryStorage())
+                )
             );
-        } catch (final IOException ex) {
+            if (!SystemUtils.IS_OS_WINDOWS) {
+                storages.add(
```
g4s8, line 128: I think we need to check if etcd storage works fine with `SubStorage` (like other storahes are checked): `new SubStorage(someKey, etcdStorage)`

g4s8 (changes requested):
@baudoliver7 looks good, just one comment and we can merge

src/test/java/com/artipie/asto/StorageExtension.java
baudoliver7: I see

baudoliver7:
@g4s8 WDYT of my last changes ?

src/main/java/com/artipie/asto/SubStorage.java
```diff
@@ -53,8 +53,14 @@ public SubStorage(final Key prefix, final Storage origin) {
 
     @Override
     public CompletableFuture<Collection<Key>> list(final Key filter) {
+        final Key cleanfilter;
+        if (this.prefix == Key.ROOT && filter == Key.ROOT) {
+            cleanfilter = Key.ROOT;
+        } else {
+            cleanfilter = new PrefixedKed(this.prefix, filter);
+        }
```
g4s8, line 61: What is a reason for these changes?

g4s8 (changes requested):
@baudoliver7 please check my question

src/main/java/com/artipie/asto/SubStorage.java
baudoliver7: It's to avoid exception :
```cmd
io.grpc.StatusRuntimeException: INVALID_ARGUMENT: etcdserver: key is not provided
```
while trying to list `ROOT` key of an Etcd's `SubStorage` with `ROOT` as prefix.
We got this exception because `new PrefixedKed(Key.ROOT, Key.ROOT)` is not equal to `Key.ROOT` as expected here (in `EtcdStorage`) :
```java
@Override
public CompletableFuture<Collection<Key>> list(final Key prefix) {
    final CompletableFuture<GetResponse> future;
    if (prefix == Key.ROOT) {
        future = this.client.getKVClient().get(
            EtcdStorage.ETCD_ROOT_KEY,
            GetOption.newBuilder()
                .withKeysOnly(true)
                .withSortOrder(SortOrder.ASCEND)
                .withRange(EtcdStorage.ETCD_ROOT_KEY)
                .build()
        );
    } else {
        future = this.client.getKVClient().get(
```
Perhaps, it would be better to do like this :
```java
if (prefix.equals(Key.ROOT))
```
and implements method `equals` of `Key`.

g4s8: @baudoliver7 yes, I think it's a good direction to make keys equitable & comparable by its values, maybe event better to use `interface Key extends Comparable<Key>` and use its `parts()` method to compare

baudoliver7: Ok, but while looking at `Key.From`, I noticed that methods `equals` and `hashCode` have already been implemented :
```java
@Override
public boolean equals(final Object another) {
    if (this == another) {
        return true;
    }
    if (another == null || getClass() != another.getClass()) {
        return false;
    }
    final From from = (From) another;
    return Objects.equals(this.parts, from.parts);
}

@Override
public int hashCode() {
    return Objects.hash(this.parts);
}
```
but decorated classes don't use them because `Key.Wrap` doesn't override them.

I will override them and make `Key` comparable also. Are you okay with ?

baudoliver7: @g4s8 Done in my last changes.

g4s8 (approved):
@baudoliver7 thanks

PR #379, opened 2021-11-22
fix: prevent file to be written out of FileStorage scope

### LINKED ISSUE #332: Prevent arbitrary file write in `FileStorage`
We should not allow to escape the sandbox of `FileStorage` and `SubStorage`, e.g. if we use `/var/artipie` as a primary storage and somebody is trying to write the key `/../../etc/passwd` the `FilesStorage` should throw and exception. Generally, the `Key` may not contain any up command and normalized path from `Key` should be always point to some sub-element.

Similar issue in RPM: https://github.com/artipie/rpm-adapter/pull/437/

baudoliver7:
Closes #332

Throw an exception when attempting to write out of `FileStorage` scope

### DIFF
--- a/src/main/java/com/artipie/asto/fs/FileStorage.java
+++ b/src/main/java/com/artipie/asto/fs/FileStorage.java
@@ -125,6 +125,10 @@ public final class FileStorage implements Storage {
 
     @Override
     public CompletableFuture<Void> save(final Key key, final Content content) {
+        final Path next = this.dir.resolve(key.string());
+        if (!next.normalize().startsWith(next)) {
+            throw new IllegalStateException("Entry path is out of storage.");
+        }
         return CompletableFuture.supplyAsync(
             () -> {
                 final Path tmp = Paths.get(
--- a/src/main/java/com/artipie/asto/fs/FileStorage.java
+++ b/src/main/java/com/artipie/asto/fs/FileStorage.java
@@ -5,14 +5,7 @@
 package com.artipie.asto.fs;
 
 import com.artipie.ArtipieException;
-import com.artipie.asto.ArtipieIOException;
-import com.artipie.asto.Content;
-import com.artipie.asto.Key;
-import com.artipie.asto.Meta;
-import com.artipie.asto.OneTimePublisher;
-import com.artipie.asto.Storage;
-import com.artipie.asto.UnderLockOperation;
-import com.artipie.asto.ValueNotFoundException;
+import com.artipie.asto.*;
 import com.artipie.asto.ext.CompletableFutureSupport;
 import com.artipie.asto.lock.storage.StorageLock;
 import com.jcabi.log.Logger;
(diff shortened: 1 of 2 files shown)


baudoliver7:
@artipie/maintainers ready to be reviewed :)

src/main/java/com/artipie/asto/fs/FileStorage.java
```diff
@@ -125,6 +125,10 @@ public FileStorage(final Path path) {
 
     @Override
     public CompletableFuture<Void> save(final Key key, final Content content) {
+        final Path next = this.dir.resolve(key.string());
+        if (!next.normalize().startsWith(next)) {
+            throw new IllegalStateException("Entry path is out of storage.");
+        }
```
genryxy, line 131: I'd suggest to wrap this exception in `ArtipieException` or just throw `ArtipieException`

genryxy (changes requested):
@baudoliver7 thanks, one suggestion

src/main/java/com/artipie/asto/fs/FileStorage.java
baudoliver7: @genryxy Ok, in this cas let throw `ArtipieException`

genryxy (approved):
@baudoliver7 thanks, looks good to me
@g4s8 please take a look

src/main/java/com/artipie/asto/fs/FileStorage.java
```diff
@@ -125,6 +125,10 @@ public FileStorage(final Path path) {
 
     @Override
     public CompletableFuture<Void> save(final Key key, final Content content) {
+        final Path next = this.dir.resolve(key.string());
+        if (!next.normalize().startsWith(next)) {
+            throw new ArtipieException("Entry path is out of storage.");
```
g4s8, line 130: This exception should be put as the result of `CompletableFuture`, you can use helper object for Java-1.8 [FailedCompletionStage](https://github.com/artipie/asto/blob/master/src/main/java/com/artipie/asto/FailedCompletionStage.java):
```java
return new FailedCompletionStage<>(new ArtipieException("Entry path is out of storage.")).toCompletableFuture();
```

Also, I'd suggest using [ArtipieIOException](https://github.com/artipie/asto/blob/master/src/main/java/com/artipie/asto/ArtipieIOException.java) here (since it's IO operation error).

g4s8 (changes requested):
@baudoliver7 thanks, please check my comment regarding code. Another comment is that `save` is not the only write operation, we also have `delete` and `move` operations which can write data to storage, so these two operations also need the same path check.

src/main/java/com/artipie/asto/fs/FileStorage.java
baudoliver7: @g4s8 got it

baudoliver7:
> @baudoliver7 thanks, please check my comment regarding code. Another comment is that `save` is not the only write operation, we also have `delete` and `move` operations which can write data to storage, so these two operations also need the same path check.

@g4s8 You are right. `delete` and `move` are another way to write write storage.

baudoliver7:
@g4s8 please, could you review the new changes ?

baudoliver7:
@g4s8 ping

baudoliver7:
@g4s8 how about my last changes ?

baudoliver7:
@g4s8 ping

baudoliver7:
@g4s8 Almost a week to wait a check/merge !

g4s8 (approved):
@baudoliver7 thanks

PR #382, opened 2021-11-25
fix: fix `deleteAll` operation

genryxy:
Part of #348
Items should be deleted from storage sequentially because otherwise some empty parts could be deleted concurrently (e.g. problems with method `FileStorage#deleteEmptyParts`)


olenagerasimova (approved):
@genryxy thanks, I guess this method will be optimized when we implement it for each storage

genryxy:
@g4s8 please, check. I corrected as we discussed

PR #385, opened 2021-12-17
fix(fs): check path for FS read

g4s8:
Refactor key path validation to use futures.
Check read key path before performing content reading.

Ref: #332


olenagerasimova (approved):
@g4s8 thanks, looks good

PR #395, opened 2022-01-24
fix(storage): add metadata method to standard decorator

baudoliver7:
We missed to add method `metadata` in `Storage.Wrap`. Here, we fix it.

Fix: #392

### DIFF
--- a/src/main/java/com/artipie/asto/Storage.java
+++ b/src/main/java/com/artipie/asto/Storage.java
@@ -199,5 +199,10 @@ public interface Storage {
         ) {
             return this.delegate.exclusively(key, operation);
         }
+
+        @Override
+        public final CompletableFuture<? extends Meta> metadata(final Key key) {
+            return this.delegate.metadata(key);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


baudoliver7:
@artipie/maintainers ready to be reviewed :)

baudoliver7:
@artipie/maintainers needs review here :)

baudoliver7:
@artipie/maintainers needs review here :)

src/main/java/com/artipie/asto/Storage.java
```diff
@@ -199,5 +199,10 @@ protected Wrap(final Storage delegate) {
         ) {
             return this.delegate.exclusively(key, operation);
         }
+
+        @Override
+        public final CompletableFuture<? extends Meta> metadata(final Key key) {
```
g4s8, line 204: We can cover this with a test

g4s8 (changes requested):
@baudoliver7 please check my comment

src/main/java/com/artipie/asto/Storage.java
baudoliver7: Interface `Storage.Wrap` is not extended by a class in this repo. All `Storage` sub-classes directly implement `Storage`.

g4s8: I think we need some kind of consistency test to be sure that `Wrap` implements all of the methods declared in `Storage` interface, e.g. `class verify extends Storage.Wrap {}` in some test will fail to compile if we update `Storage` interface but forget to change its `Wrap` implementation.

baudoliver7: I see now

g4s8 (approved):
@baudoliver7 thanks!

PR #402, opened 2022-02-18
fix: change way we check that key path is contained in `FileStorage` location

baudoliver7:
We noticed that when the key path contains a `.` element, we get an exception saying that the entry path is not in storage location.
We refactored `FileStorage` by introducing a class `KeyPath` to easily fix the bug.

Cc: artipie/artipie#995

### DIFF
--- a/src/main/java/com/artipie/asto/KeyPath.java
+++ b/src/main/java/com/artipie/asto/KeyPath.java
@@ -39,7 +39,7 @@ public final class KeyPath {
      * @param key Key to transform
      */
     public KeyPath(final Path root, final Key key) {
-        this.root = root;
+        this.root = root.normalize();
         this.key = key;
     }
 
@@ -53,13 +53,12 @@ public final class KeyPath {
      * @return Path future
      */
     public Path get() {
-        final Path path = this.root.resolve(this.key.string());
-        final Path pathn = path.normalize();
-        if (pathn.startsWith(this.root)) {
-            return pathn;
+        final Path path = this.root.resolve(this.key.string()).normalize();
+        if (path.startsWith(this.root)) {
+            return path;
         } else {
             throw new ArtipieIOException(
-                String.format("Entry path is out of root location: %s", pathn)
+                String.format("Entry path is out of root location: %s", path)
             );
         }
     }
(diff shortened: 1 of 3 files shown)


baudoliver7:
@artipie/maintainers please, needs review :)
@g4s8 besides, could you please, mark `pdd-lint` as required ?

baudoliver7:
@g4s8 I agree with you. But when retrieving key path in `FileStorage`, we used this code before :
```java
private CompletableFuture<? extends Path> keyPath(final Key key) {
    final Path path = this.dir.resolve(key.string());
    final CompletableFuture<Path> res = new CompletableFuture<>();
    if (path.normalize().startsWith(path)) {
        res.complete(path);
    } else {
        res.completeExceptionally(
            new ArtipieIOException(
                String.format("Entry path is out of storage: %s", key)
            )
        );
    }
    return res;
}
```
You will notice that when `path` variable contains a `.` element, an exception is thrown because `path.normalize()` ignores it.
So I changed the way we check that the path is contained in storage location. That's what I call support `.` element :)

baudoliver7:
@g4s8 Here is the new way to check :

```java
public Path get() {
    final Path path = this.root.resolve(this.key.string());
    final Path pathn = path.normalize();
    if (pathn.startsWith(this.root)) {
        return pathn;
    } else {
        throw new ArtipieIOException(
            String.format("Entry path is out of root location: %s", pathn)
        );
    }
}
```
I think it would be better to normalize also `this.root` while passing it in `KeyPath` constructor.

g4s8:
> > @baudoliver7 I'm not sure if it's a valid fix for the problem: dot symbol is not a valid key part (e.g. for S3 or etcd storage). Also, we intentionally prevent any "relativeness" in keys to avoid sandbox escaping or other bugs. So I think the correct fix here is to fix the key which causes such an exception.
> 
> @g4s8 I think we should retrieve key path as long as it is a valid path and contained in the storage location. Besides, dot symbol could be in the `Storage` location and we will have the same trouble. IMHO, it would be better to take account of them when checking that key path is in `Storage` location or not.
> 
> I added some little changes for the last remark.

@baudoliver7 the problem I'm talking about is that the key is not a path. Each valid key should be supported by all storage, not only by some specific, and we should interpret these keys equal for all types of storage. E.g. `S3Storage.value(key) and `FileStorage.value(key)` should produce the same result for key `foo/./bar`.

baudoliver7:
@g4s8 Ok. Let's then fix it in [conda-adapter](https://github.com/artipie/conda-adapter).

Issue #413, opened 2022-04-22
Fix test for `LoggingStorage`


olenagerasimova:
Let's fix `LoggingStorageTest`: after logging library updates, logger does not write into writer, configured in the test. Figure out how to deal with it and enable the test.

baudoliver7:
@olenagerasimova hi, with pleasure ! :)

olenagerasimova:
@baudoliver7 nice, waiting for PR :)

baudoliver7:
@olenagerasimova PR submitted [here ](https://github.com/artipie/asto/pull/420):)

PR #420, opened 2022-05-16
test: fix tests for `LoggingStorage`

baudoliver7:
`jcabi-log:0.20.1` does not support recent versions of logging libraries.
So, we replaced it by `java.util.logging`.

Close: #413


olenagerasimova (approved):
@baudoliver7 thanks, merging

Issue #421, opened 2022-05-18
Wrong implementation of the S3Storage#putMultipart


dgarus:
For each uploaded part of a long object, the implementation must provide the part number and the `ETag` value, 
returned after that part was uploaded (see [CompleteMultipartUpload ](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html)).
But currently this is not done, so multipart downloading is not working.
Related test is `com.artipie.asto.s3.S3StorageTest#shouldUploadObjectWhenSaveLargeContent`.

PR #425, opened 2022-05-20
fix: added parts information to `CompleteMultipartUploadRequest`

dgarus:
closes: #421

- created `Splitting` class to split `ByteBuffer` at parts
- created tests to check `Splitting` implementation
- added filling of `multipartUpload` field of `CompleteMultipartUploadRequest`

### DIFF
--- a/src/main/java/com/artipie/asto/Splitting.java
+++ b/src/main/java/com/artipie/asto/Splitting.java
@@ -49,7 +49,7 @@ public class Splitting {
         final Publisher<ByteBuffer> res;
         int remaining = this.source.remaining();
         if (remaining > this.size) {
-            final List<ByteBuffer> parts = new ArrayList<>(16);
+            final List<ByteBuffer> parts = new ArrayList<>(remaining / this.size + 1);
             while (remaining > 0) {
                 final byte[] bytes;
                 if (remaining > this.size) {
--- a/src/main/java/com/artipie/asto/Splitting.java
+++ b/src/main/java/com/artipie/asto/Splitting.java
@@ -1,0 +1,70 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/asto/LICENSE.txt
+ */
+package com.artipie.asto;
+
+import io.reactivex.Flowable;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+import org.reactivestreams.Publisher;
+
+/**
+ * Splits the original ByteBuffer to several ones
+ * with size less or equals defined max size.
+ *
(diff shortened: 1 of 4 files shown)


dgarus:
@olenagerasimova 
Could you please have a look?

src/main/java/com/artipie/asto/Splitting.java
```diff
@@ -0,0 +1,70 @@
...
+     * Splits the original ByteBuffer to ones with size less
+     * or equals defined max {@code size}.
+     *
+     * @return Publisher of ByteBuffers.
+     */
+    public Publisher<ByteBuffer> publisher() {
+        final Publisher<ByteBuffer> res;
+        int remaining = this.source.remaining();
+        if (remaining > this.size) {
+            final List<ByteBuffer> parts = new ArrayList<>(16);
```
olenagerasimova, line 52: why 16? It seems, like we can calculate the size by dividing `remaining` into `this.size`

src/main/java/com/artipie/asto/s3/MultipartUpload.java
```diff
@@ -70,36 +83,33 @@ final class MultipartUpload {
...
+                buffer -> Flowable.fromPublisher(
+                    new Splitting(buffer, MultipartUpload.MIN_PART_SIZE).publisher()
+                )
+            ).map(
+                chunk -> {
+                    final int pnum = counter.incrementAndGet();
+                    return MultipartUpload.this.uploadPart(
+                        pnum,
+                        Flowable.just(chunk)
+                    ).<Void>thenApply(
```
olenagerasimova, line 98: It would be better to use `thenAccept` here instead of  `thenApply` as `thenAccept` returns void itself

src/main/java/com/artipie/asto/s3/MultipartUpload.java
```diff
@@ -70,36 +83,33 @@ final class MultipartUpload {
...
-            ).<Void>thenApply(ignored -> null)
-        ).reduce(
-            CompletableFuture.allOf(),
-            (acc, stage) -> acc.thenCompose(o -> stage)
-        ).to(SingleInterop.get()).toCompletableFuture().thenCompose(Function.identity());
+            ).reduce(
+                CompletableFuture.allOf(),
+                (acc, stage) -> acc.thenCompose(o -> stage)
+            ).to(SingleInterop.get())
+            .toCompletableFuture()
```
olenagerasimova, line 111: This line is not required, we already switched to completables, `thenCompose` can be called without calling `toCompletableFuture()`

olenagerasimova (changes requested):
@dgarus thanks, check my comments please

olenagerasimova (approved):
@dgarus thanks, looks good

Issue #427, opened 2022-05-27
`StorageValuePipeline#processWithResult` is hanging in the case when value is getting by key "from" has large size


dgarus:
`StorageValuePipeline#processWithResult` is hanging when reading storage data size more than 1024 bit (default size of `PipeInputStream` buffer).

PR #428, opened 2022-05-27
fix: StorageValuePipeline#processWithResult hanging

dgarus:
Close: #427
 - implemented asynchronous starting of read, action and write futures inside the `StorageValuePipeline#processWithResult` method.
 - added test to check `StorageValuePipeline` in the case when value is getting by key "from" has large size.


dgarus:
@olenagerasimova hello!
Could you please do the review?

olenagerasimova (approved):
@dgarus thanks, code looks good, merging. I have some questions though, it would be more convenient to discuss them in the office on Monday.

PR #460, opened 2022-08-15
Remove `wtf.g4s8:tuples` dependency

dgarus:
closes: #459 

- removed dependency
- changed `Pair` to `ImmutablePair`


olenagerasimova (approved):
@dgarus thanks

Issue #445, opened 2022-07-19
feat: create test framework for testing new storage implementations 


dgarus:
The module `asto-core` contains many universal storage tests `StorageExtension`.
We should allow using this tests to test new storage implementations.

We can create a new module `asto-test`, move common test classes to this module and 
implement something like [Reactive Streams TCK](https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck).

- [x] move common storage tests to src (temporary, later it should be in a separate module)
- [x] move `S3Storage` to separate module
- [x] move `EtcdStorage` to separate module
- [x] move `VertxFileStorage` to separate module
- [x] fix `VertxFileStorage` verification tests for case root sub storage
- [x] add verification test to asto-redis
- [x] move `ArtipieStorage` to separate module
- [x] document how to use verification tests

PR #468, opened 2022-09-01
VertxFileStorage as separate module

dgarus:
part of #445 

Moved `VertxFileStorage` to a separate module


olenagerasimova (approved):
@dgarus thanks

PR #469, opened 2022-09-02
Move EtcdStorage to a separate module

dgarus:
part of #445

Moved `EtcdStorage` to a separate module.


olenagerasimova (approved):
@dgarus thanks

PR #472, opened 2022-09-07
RedisStorage verification tests

dgarus:
part of #445 

- fixed an error in verification tests
- added `RedisStorage` verification tests


olenagerasimova (approved):
@dgarus thanks

PR #473, opened 2022-09-09
fix `VertxFileStorage` verification tests for case root sub storage

dgarus:
part of #445

Fixed VertxFileStorage verification tests for case root sub storage

### DIFF
--- a/asto-core/src/main/java/com/artipie/asto/Key.java
+++ b/asto-core/src/main/java/com/artipie/asto/Key.java
@@ -54,10 +54,38 @@ public interface Key {
     List<String> parts();
 
     /**
+     * Base key class.
+     * @since 1.14.0
+     */
+    abstract class BaseKey implements Key {
+        @Override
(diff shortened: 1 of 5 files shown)


asto-core/src/main/java/com/artipie/asto/Key.java
```diff
@@ -53,11 +53,39 @@ public interface Key {
      */
     List<String> parts();
 
+    /**
+     * Base key class.
+     * @since 1.14.0
+     */
+    abstract class BaseKey implements Key {
```
olenagerasimova, line 60: I'd suggest to call this class just `Base` without `key` as this class is a part of `Key` interface and will always be accessed as `Key.Base`.

dgarus: ok

olenagerasimova (approved):
@dgarus thanks

dgarus:
@olenagerasimova thank you for the review!

PR #474, opened 2022-09-12
move `ArtipieStorage` to a separate module

dgarus:
part of #445 

Moved `ArtipieStorage` to a separate module.


asto-artipie/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+    <dependency>
+      <groupId>com.artipie</groupId>
+      <artifactId>asto-core</artifactId>
+      <version>1.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.artipie</groupId>
+      <artifactId>http-client</artifactId>
+      <version>0.3.6</version>
```
olenagerasimova, line 43: use latest http-client version, it does not depend on asto

olenagerasimova (changes requested):
@dgarus thanks, check my comments please
But it seems to me we talked about moving this storage to files-adapter in order to avoid dependencies cycle

dgarus:
@olenagerasimova thanks for review!
The problem with cyclic dependencies is solved. `asto-core` doesn't have reference on http-client and http-client doesn't have reference on asto-artipie.

PR #479, opened 2022-10-14
fix: verification base tests corrected and file storage fixed

olenagerasimova:
Whitebox Verification base test should use junit jupiter `Test` annotation to actually run the tests for various storages and I've fixed error in `FileStorage.list()` method (we could not have missed it if Verification test was actually working).


olenagerasimova:
@dgarus please, review, file storage is broken in current release built, let's merge the fix and release correct version

dgarus (approved):
@olenagerasimova thanks!

PR #481, opened 2022-11-22
feat: storage identifier

### LINKED ISSUE #480: Storage identifier for monitoring
Let's add default method to storage interface, which will provide some unique information about the storage instance.

olenagerasimova:
Closes #480

Added default method to get some storage identifier into `Storage` interface and overrode it where required.


dgarus (changes requested):
@olenagerasimova thanks!

A common comment: an identifier value should be calculated only once.
Therefore, it has to be a `private final` field.

dgarus (approved):
Thanks! LGTM

PR #488, opened 2023-01-19
feat: abstract class to load factories

### LINKED ISSUE #487: Factories loader
We are adding more factories for various objects and classes to load them are very much alike. So, let's extract abstract class from `Storages` and make the `Storages` class to extend it. Also, let's rename `StorageConfig` into `Config` to make available to reuse.

olenagerasimova:
closes #487 

check the issue for more details


dgarus (approved):
Thanks! lgtm

PR #500, opened 2023-03-23
feat: set of default packages in `FactoryLoader`

### LINKED ISSUE #499: Default packages in factory loader
Let's make `FactoryLoader.defPackage()` return set of default packages as now we need two default packages for permissions: http and docker-adapter.

olenagerasimova:
closes #499


andpopov (approved):
@olenagerasimova Good idea, thanks!

dgarus (approved):
Thanks!

PR #501, opened 2023-04-26
S3 credentials should be optional

### LINKED ISSUE #1281: S3 credentials should be optional
The current S3 storage config requires API Key and Secret.
```
bucket (string, required) - bucket name
region (string, optional) - bucket region name
endpoint (string, optional) - S3 API provider URL, default is standard AWS S3 endpoint
credentials (map, required):
type (string, required) - authentication type, one of: basic
accessKeyId (string, required) - access API key ID
secretAccessKey (string, required) - secret key
```
 
This is not a good practice.
AWS SDKs are designed to read credentials from multiple sources, (ENV, metadata URL(EC2,ECS..)) and therefore it is no need to provide credentials directly. 
Hence this should be optional and let AWS SDK handle it for you. So that starting the server would only require `AWS_` env vars in place.

dgarus:
Close https://github.com/artipie/artipie/issues/1281

- credentials are made optional for S3 storage factory; added test;
- added using S3 storage factory to get a storage instance in tests.

### DIFF
--- a/asto-core/src/main/java/com/artipie/asto/factory/Config.java
+++ b/asto-core/src/main/java/com/artipie/asto/factory/Config.java
@@ -43,6 +43,13 @@ public interface Config {
     Config config(String key);
 
     /**
+     * Checks that there is a config data.
+     *
+     * @return True if no config data.
+     */
+    boolean isEmpty();
+
+    /**
      * Strict storage config throws {@code NullPointerException} when value is not exist.
      *
      * @since 1.13.0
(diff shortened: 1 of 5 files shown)


olenagerasimova (approved):
@dgarus thanks, also, I think, we need to update wiki by specifying that `credentials` in s3 config are optional and can be set in env variables.
And, as `Config` interface was changed, `http` has to be updated and this new method should be implemented.

Issue #502, opened 2023-05-19
Events processing mechanism


olenagerasimova:
Let's add some general mechanism to collect and process some data. The data are some information from the adapters about the package or user action, so it should be possible to pass this data into some in-memory cache or queue in no time. 
Then, the data should be somehow processed by separate threads/mechanisms.

olenagerasimova:
will be implemented in artipie/artipie repo

PR #503, opened 2023-05-19
feat: added quartz and scheduled jobs

olenagerasimova:
part of #502 
Added skeleton for periodic processing of some data from queue. The idea is to use it for gathering and processing some events data from adapters. Implementation is based on [quartz scheduler](http://www.quartz-scheduler.org/). Quartz is strong open source scheduler, support clustering and a lot of other features. 
- `QuartsService` class is the main class to start scheduling and add some jobs
- `EventsProcessor` is quartz job to process pack of elements from queue using provided consumer
- `EventQueue` is the queue to gather data for processing

Data can be any object, for example, info about added package (name, version, owner atc). Consumer knows what to do with the pack of data, for example, it can write data to db.

### DIFF
--- a/asto-core/src/main/java/com/artipie/asto/events/EventQueue.java
+++ b/asto-core/src/main/java/com/artipie/asto/events/EventQueue.java
@@ -1,0 +1,48 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2023 artipie.com
+ * https://github.com/artipie/asto/LICENSE.txt
+ */
+package com.artipie.asto.events;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * Events queue with {@link ConcurrentLinkedQueue} under the hood.
+ * Instance of this class can be passed where necessary (to the adapters for example)
+ * to add data for processing into the queue.
+ * @param <T> Queue item parameter type.
+ * @since 1.17
+ */
+public final class EventQueue<T> {
+
+    /**
+     * Queue.
+     */
+    @SuppressWarnings("PMD.AvoidFieldNameMatchingMethodName")
+    private final Queue<T> queue;
+
+    /**
+     * Ctor.
+     */
+    public EventQueue() {
+        this.queue = new ConcurrentLinkedQueue<>();
+    }
+
+    /**
+     * Add item to queue.
+     * @param item Element to add
+     */
+    public void put(final T item) {
+        this.queue.add(item);
+    }
+
+    /**
+     * Queue, not public intentionally, the queue should be accessible only from this package.
+     * @return The queue.
+     */
+    @SuppressWarnings("PMD.AvoidFieldNameMatchingMethodName")
+    Queue<T> queue() {
+        return this.queue;
+    }
+}
--- a/asto-core/src/main/java/com/artipie/asto/events/QuartsService.java
+++ b/asto-core/src/main/java/com/artipie/asto/events/QuartsService.java
@@ -6,7 +6,7 @@ package com.artipie.asto.events;
 
 import com.artipie.ArtipieException;
 import com.jcabi.log.Logger;
-import java.util.Collection;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.function.Consumer;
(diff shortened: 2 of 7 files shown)


olenagerasimova:
@dgarus please, take a look, let's discuss if necessary

asto-core/src/main/java/com/artipie/asto/events/EventQueue.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * Events queue with {@link ConcurrentLinkedQueue} under the hood.
+ * Instance of this class can be passed where necessary (to the adapters for example)
+ * to add data for processing into the queue.
+ * @param <T> Queue item parameter type.
+ * @since 1.17
+ */
+public final class EventQueue<T> {
```
dgarus, line 17: Why do we need this wrapper? 
Can we use `ConcurrentLinkedQueue` directly?

asto-core/src/main/java/com/artipie/asto/events/QuartsService.java
```diff
@@ -0,0 +1,96 @@
...
+     * given seconds.
+     * @param queue Queue to get data from
+     * @param consumer How to consume the data collection
+     * @param seconds Seconds interval for scheduling
+     * @param <T> Data item object type
+     * @throws SchedulerException On error
+     */
+    public <T> void addPeriodicEventsProcessor(
+        final EventQueue<T> queue, final Consumer<Collection<T>> consumer, final int seconds
+    ) throws SchedulerException {
```
dgarus, line 69: What do we get useful for us if we will use `quarts` rather than the standard `ScheduledExecutorService`?

asto-core/src/main/java/com/artipie/asto/events/EventQueue.java
olenagerasimova: To hide the queue from being modified (cleaned or polled) from anywhere except this package. Only adding elements should be allowed from outside.

asto-core/src/main/java/com/artipie/asto/events/QuartsService.java
olenagerasimova: It's mush more powerful, for example, `ScheduledExecutorService` allows only to perform some periodical jobs. When we add some workers to match/check repository data and our db, we will need to schedule them for every night (for example). `ScheduledExecutorService` does not allow to do it out of the box, we will need `quarts`

olenagerasimova:
@dgarus I've checked all quartz docs and now I'm sure that there is no way to directly specify number of threads per job. 

> if you have a mix of light-weight and heavy-weight jobs, then you may wish to have a scheduler with many threads to service the lightweight jobs and one with few threads to service the heavy-weight jobs, in order to keep your machines resources from being overwhelmed by running to many heavy-weight jobs concurrently 

Says this [example page](https://github.com/quartz-scheduler/quartz-scheduler.org-site/blob/master/documentation/quartz-2.3.0/cookbook/MultipleSchedulers.md). 

Quartz scheduler can be instantiated from config file (so, app can have several schedulers created from different configs), we can win from such configuration by providing users an opportunity to optionally specify config file and thus create another scheduler per custom job.

dgarus:
> we can win from such configuration by providing users an opportunity to optionally specify config file and thus create another scheduler per custom job.

It sounds well.

olenagerasimova:
@dgarus updated code. The main changes:
- schedule required amount of jobs to run in parallel. The jobs share the same job data and thus process the same queue with given consumer
- no collection inside consumer
- added setters which are called to deliver queue and consumer to the job from job data by quartz

olenagerasimova:
As for the monitoring: 
1) there is [quartz desk](https://www.quartzdesk.com/), free version also avaliable
> QuartzDesk is a highly flexible enterprise-class management and monitoring platform for Quartz schedulers embedded in Java applications

2) we can use [listeners](https://github.com/quartz-scheduler/quartz/blob/main/docs/tutorials/tutorial-lesson-07.md) and add micrometer and jfr events manually. Also, queue monitoring should be added manually in artipie.

PR #504, opened 2023-12-01
Trying to hot-fix s3 bug. Hangs in reads from ContentAsInputStream

ChGen:
Trying to hot-fix s3 bug. Hangs in reads from `ContentAsInputStream.inputStream()`.
Instances of `PublisherAs(content)` can read data from content correctly, so two new methods were introduced, which use `byte[]` instead of `InputStream`.

### DIFF
--- a/asto-core/src/main/java/com/artipie/asto/streams/StorageValuePipeline.java
+++ b/asto-core/src/main/java/com/artipie/asto/streams/StorageValuePipeline.java
@@ -9,6 +9,7 @@ import com.artipie.asto.ByteArray;
 import com.artipie.asto.Content;
 import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
+import com.artipie.asto.ext.PublisherAs;
 import com.artipie.asto.misc.UncheckedIOConsumer;
 import com.artipie.asto.misc.UncheckedIOSupplier;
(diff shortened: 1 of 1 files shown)


ChGen:
@dgarus , please review/comment. This is more like a hot-fix, but it passes all tests locally for asto, conda-adapter & artipie.

olenagerasimova:
@ChGen I've checked `com.artipie.conda.meta.MergedJson.Jackson` class and method `merge()` and noticed, that `JsonParser` is not closed anywhere in this method. Have you tried to call `close()`? 

As for this approach... The whole idea of this class is NOT TO READ PUBLISHER INTO BYTE ARRAY, so I do not agree with adding these methods into asto. Fix for conda can be done inside conda via `PublisherAs(content).getBytes()`, no need to add these strange methods into asto

ChGen:
@olenagerasimova , thanks, I'll try your suggestions. I understand that reading all to byte array isn't good anyway, but I'd rather to publish working hotfix in predictable manner, and than debug this properly as a next step.

ChGen:
Actually, metadata is small, compared to the artifacts, so we are quite safe with that.

ChGen:
Hotfix provided in https://github.com/artipie/artipie/pull/1351

